PR #68, opened 2018-01-24
Control shutdown of router with custom script

blackfuel:
### Controlled shutdown of router with custom script
This patch allows the user to define a custom script named `shutdown-start`, and optionally cancel a reboot or halt.  I've hooked it into most places you expect it to be, like the WebUI, rc , init and watchdog.  I just need better control of the shutdown, so I don't lose any work/data.  I verified it works with both RT-AC68U and RT-AC86U routers.  Just looking for people to help me finish testing this.

### Example
**/jffs/scripts/shutdown-start**
```
#!/bin/sh
#nvram set shutdown_cancel=1  # for testing purposes

wall_write() {
  logger -t $(basename $0) "$1"
  for tty in /dev/pts/*; do
    if [ -c $tty ]; then
      echo "$1" >$tty
    fi
  done
}

wall() {
  if [ -n "$1" ]; then
    wall_write "$1"
  else
    local textline=""
    while read -r textline; do
      wall_write "$textline"
    done
  fi
}

wall "The system is going down in 10 seconds.  Please log off now or you may lose some work."
sleep 10
echo "Bye bye." | wall
sleep 1
```

### Testing the feature
- WebUI/browser: press the Reboot button
    OR
- Shell command: `reboot`
    OR
- Shell command: `kill -TERM 1`

### DIFF
--- a/release/src/router/httpd/web.c
+++ b/release/src/router/httpd/web.c
@@ -3740,6 +3740,7 @@ static int ej_set_variables(int eid, webs_t wp, int argc, char_t **argv) {
 		}
 	}
 	else if(!strcmp(apiName, "qos")){
+		if (shutdown_start("httpd", __FUNCTION__, apiName, apiAction) == 0) {
 		char *desc, *addr, *port, *prio, *transferred, *proto;
 
 		if(!strcmp(apiAction, "enable")){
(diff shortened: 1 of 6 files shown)


RMerl:
I find the patch too invasive, as it branches all over the code, and would require to be manually kept up-to-date whenever Asus adds new service events that might be related.

The new service-event script added in 384.5 is probably a simpler way to handle it, by looking for "reboot" event types.

PR #149, opened 2018-05-19
kernel: enable NETEM for tc

xuecai:


RMerl:
Any particular reason to want this as part of the base firmware?  Doesn't look like something anyone else would need.

xuecai:
In development, i need to use TC and netcm modules to simulate a weak network environment。
For more developers, in Merlin firmware, to lower the barriers to use when using weak-net-like simulation software like Atc。

robcore:
Hi, Eric.
I'm Rob, and relatively new to building router firmware,  but have been involved in linux development for years (desktop and embedded android garbage).  I am actually in the process of including sch_netem for a personal build of your awesome fork due to a need for some testing I'm doing as well.  If it is successful, I will be sure to let you know, but I wanted to confirm that the interest exists for more than one person.  Take care!
- Rob

PR #153, opened 2018-05-26
changed namecheap ddns url

ShadowNoire:
park-your-domain has been having auth issues but dnsomatic seems to work fine


RMerl:
dns-o-matic is a completely different provider (OpenDNS).

PetNoire:
it may be different but works for both my namecheap domains

PR #194, opened 2018-09-05
uses 64bit int instead of pointer in xt_geoip related structs, to fix user space and kernel space struct size mismatch

SylphiaWindy:
Actually only these models had 64bit kernel but 32bit user space would encounter this problem.

hnd toolchain must also being modified, too, as extensions of iptables was built against headers in toolchain.


paldier:
what is the problem

SylphiaWindy:
in short, iptables geoip module not working on rt-ac86u with default kernel

RMerl:
Closing, as I have no plan to merge this.  Unnecessary toolchain changes for a component that is not used by the firmware.

PR #220, opened 2018-10-23
Fix several broken click event handlers

mgziminsky:
There are a number of click handlers that are broken due to an undefined event variable.
The one that I noticed specifically prevents you from removing port trigger rules, which fails on line 1301.
I'm not exactly sure if/where the others are used, but they have the same issue code-wise, so I fixed them also.


RMerl:
Firefox might simply be more tolerant of that particular error.  I will do some tests to confirm, and if indeed fails on Firefox I will merge that change.

RMerl:
Seems to be specific to Firefox indeed:

https://stackoverflow.com/questions/20522887/referenceerror-event-is-not-defined-error-in-firefox

PR #263, opened 2019-02-10
Don't write off the end of the array.

dp111:


RMerl:
btconfig is only used by some specific Lyra models, none of which are used by Asuswrt-Merlin.  Best course of action rather is for me to remove btconfig from the tree since we don't need it.

PR #266, opened 2019-02-11
Remove double sizeof

dp111:


RMerl:
SDK component that is not used by Asuswrt.

PR #269, opened 2019-02-11
Don't risk over running the buffer

dp111:


RMerl:
A cleaner fix would be to use sizeof(com) rather than another hardcoded value, to prevent future changes from breaking it again.

PR #271, opened 2019-02-11
Make array argv the right size

dp111:


RMerl:
Going to merge this, however note that this plugin is not used by the firmware, so this fix should rather be sent upstream to the pppd developers (if it still exists).

PR #272, opened 2019-02-11
Correct range test so it isn't always true

dp111:


RMerl:
Content of the contrib folder is not used by the firmware.

PR #273, opened 2019-02-11
check what is pointed to by productList and not the pointer against '\0'

dp111:


RMerl:
Something wrong with this PR, as it changes the entire source file.

Also, please write commit entries in the following format:

subsystem: description of the change.

For example:

httpd: fixed buffer overrun in thisfunction()

This makes it easier to follow the commit log afterward.

dp111:
The issue is the file has mixed line endings and github is normalising them
.

In function : search_devices

There are the following lines :


/* Sanity check */
if (!vendor || (!product && productList == '\0') )
return NULL;

if (productList != '\0')
listcopy = malloc(strlen(productList)+1);


I think productList  should be *productList

RMerl:
Removing it, Asuswrt-Merlin uses the 2.4.0 version present in usb-modeswitch/

PR #274, opened 2019-02-12
check what p points to and not p for '\0'

dp111:
in function mschap_xlat (p== '\0') was being used instead of (*p=='\0')


themiron:
as pointed earlier, please follow commit msg scheme "package/subsystem: what was changed"
it's quite hard to navigate over one and the same commits.
thanks

dp111:
In the freeradius-server release/src/router/freeradius-server-3.0.0/src/modules/rlm_mschap/rlm_mschap.c
 The  function mschap_xlat (p== '\0') was being used instead of (*p=='\0')

RMerl:
Freeradius isn't used by Asuswrt at this time anyway (no idea why Asus has it in the source tree, maybe for the BRT line of routers).

PR #308, opened 2019-04-28
wanduck: Fix stack-based buffer-overflow in DNS handling

KiloFoxtrotPapa:
Incorrect calculation of the 'end' pointer results in buffer overwrite in
subsequent strcat and strncat calls.

Potential overflow of MAXLINE (2048) - PATHLEN (256) bytes.

Other miscellaneous fixes:
 - initialize structs preemptively
 - use sizeof() on struct just in case of misalignment when doing memset()
 - fix redundant strlen() check to the length we care about

### DIFF
--- a/release/src/router/rc/wanduck.c
+++ b/release/src/router/rc/wanduck.c
@@ -2340,8 +2340,8 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 #endif
 	unsigned char reply_content[MAXLINE], *ptr, *end;
 	dns_header *d_req, *d_reply;
-	dns_queries queries;
-	dns_answer answer;
+	dns_queries queries = {0};
+	dns_answer answer = {0};
 	uint16_t opcode;
 
 	/* validation */
@@ -2352,10 +2352,11 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 		return;
 	opcode = d_req->flag_set.flag_num & htons(0x7800);
 	ptr = request + sizeof(dns_header);
-	end = request + maxlen;
+	/* queries.name is PATHLEN long, one trailing byte for null */
+	end = request + (maxlen > PATHLEN-1 ? PATHLEN-1 : maxlen);
 
 	/* query, only first so far */
-	memset(&queries, 0, sizeof(queries));
+	memset(&queries, 0, sizeof(dns_queries));
 	while (ptr < end) {
 		size_t len = *ptr++;
 		if (len > 63 || end - ptr < (len ? : 4))
@@ -2371,7 +2372,7 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 		strncat(queries.name, (char *)ptr, len);
 		ptr += len;
 	}
-	if (queries.type == 0 || queries.ip_class == 0 || strlen(queries.name) > 1025)
+	if (queries.type == 0 || queries.ip_class == 0 || strlen(queries.name) > PATHLEN)
 		return;
 	maxlen = ptr - request;
 


RMerl:
Pinging @themiron as he's more familiar than me with this particular section of the code.

release/src/router/rc/wanduck.c
```diff
@@ -2340,8 +2340,8 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 #endif
 	unsigned char reply_content[MAXLINE], *ptr, *end;
 	dns_header *d_req, *d_reply;
-	dns_queries queries;
-	dns_answer answer;
+	dns_queries queries = {0};
```
themiron: initing with {0} is not necessary, there's memset(&queries, 0, sizeof(queries)) down below

release/src/router/rc/wanduck.c
```diff
@@ -2352,10 +2352,11 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 		return;
 	opcode = d_req->flag_set.flag_num & htons(0x7800);
 	ptr = request + sizeof(dns_header);
-	end = request + maxlen;
+	/* queries.name is PATHLEN long, one trailing byte for null */
+	end = request + (maxlen > PATHLEN-1 ? PATHLEN-1 : maxlen);
```
themiron: unrelated, dns labels don't use trailing null bytes. end pointer is the absulute end of the packet, not of single hostname or so.

release/src/router/rc/wanduck.c
```diff
@@ -2352,10 +2352,11 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
...
 
 	/* query, only first so far */
-	memset(&queries, 0, sizeof(queries));
+	memset(&queries, 0, sizeof(dns_queries));
```
themiron: quite valid, but what if sizeof(valiable) here is safer than sizeof(typeof(variable))

release/src/router/rc/wanduck.c
```diff
@@ -2340,8 +2340,8 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 #endif
 	unsigned char reply_content[MAXLINE], *ptr, *end;
 	dns_header *d_req, *d_reply;
-	dns_queries queries;
-	dns_answer answer;
+	dns_queries queries = {0};
+	dns_answer answer = {0};
```
themiron: initing with {0} is not necessary, there's memset(&answer 0, sizeof(answer)) down below

release/src/router/rc/wanduck.c
```diff
@@ -2371,7 +2372,7 @@ void handle_dns_req(int sfd, unsigned char *request, int maxlen, struct sockaddr
 		strncat(queries.name, (char *)ptr, len);
 		ptr += len;
 	}
-	if (queries.type == 0 || queries.ip_class == 0 || strlen(queries.name) > 1025)
+	if (queries.type == 0 || queries.ip_class == 0 || strlen(queries.name) > PATHLEN)
```
themiron: 1025 is max hostname length from standard, although name is PATHLEN==256 sized, better to keep it in case of PATHLEN will be increased.

themiron:
@KiloFoxtrotPapa please help to review https://github.com/RMerl/asuswrt-merlin.ng/commit/dbfbc26a18f28a9bb8934bf929d0306b0d148f64

release/src/router/rc/wanduck.c
KiloFoxtrotPapa: I would've removed the memset(), but reading the comment I assumed you planned for implementing handling of multiple queries so I left it there.

release/src/router/rc/wanduck.c
KiloFoxtrotPapa: Doing so increase the possibility of zero'ing not enough bytes if the struct contains padding. That being said, sizeof(variable) is preferable when doing arithmetic.

This is typically more significant when trying to erase sensitive data, for this project it likely won't cause an issue.

release/src/router/rc/wanduck.c
KiloFoxtrotPapa: This has less to do with DNS but more with the fact that strcat() strcpy() calls rely on the position of the null byte for their operations.

release/src/router/rc/wanduck.c
KiloFoxtrotPapa: I don't feel it's wanduck's job to validate if DNS queries conform to the RFC or not, but that's just a personal opinion.

KiloFoxtrotPapa:
> @KiloFoxtrotPapa please help to review [dbfbc26](https://github.com/RMerl/asuswrt-merlin.ng/commit/dbfbc26a18f28a9bb8934bf929d0306b0d148f64)

Commit looks good, feel free to close this PR.

release/src/router/rc/wanduck.c
themiron: since it implements dns stub, it must follow related dns protocol rules incl. validation

release/src/router/rc/wanduck.c
themiron: if sizeof of struct & variable doesn’t match there should be something very wrong with compiler. using one or another approach, I think, depends on exact case, i.e while ops on variable - preferable use sizeof(variable), while allocating struct - malloc(sizeof(struct)) although sizeof(*variable) is still valid.
as for cleaning struct members incl. possible padding , there’s no other way than to zero from this member start to next member start. not the case here, the whole struct is zeroed.
so.. matter of taste, in the end :)

release/src/router/rc/wanduck.c
themiron: sure, you’re right, but I mean a bit different thing. cutting “end” too early at pathlen-1 would cut off type/class fields placed after dns name, therefore it will efficiently break validation at all

KiloFoxtrotPapa: Ah, got it and you're right. I'd usually make it buffer[PATHLEN+1] but I didn't want to modify the .h file. But again you're right about breaking DNS parsing in this case.

release/src/router/rc/wanduck.c
KiloFoxtrotPapa: Hmm..I think you're right in that sizeof(variable) is safer. We might later change the type.
I'm probably thinking about sizeof(variable) not necessarily equals sizeof(variable.a) + sizeof(variable.b)...etc.

PR #309, opened 2019-04-30
Removing the duplicate 36/40/44 channel option

AlfnXd:
Removing the duplicate 36/40/44 channel option when selecting bandwidth of 80MHz

### DIFF
--- a/release/src/router/www/sysdep/RT-AC66U/www/chanspec.js
+++ b/release/src/router/www/sysdep/RT-AC66U/www/chanspec.js
@@ -136,7 +136,7 @@ function wl_chanspec_list_change(){
 					for(i=0;i<wl_channel_list_5g.length; i++){
 						var _cur_channel = parseInt(wl_channel_list_5g[i]);
 						for(j=0;j<wl1.channel_80m.length;j++){
-							if(wl1.channel_80m[j].indexOf(_cur_channel) != -1){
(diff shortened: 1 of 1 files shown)


RMerl:
What router model and browser are you seeing an issue?  The original code looks good to me.  indexOf() returns the position the string is found, or -1 if it's not found.  Original code says "if the string is found anywhere within", yours simply says "if the string is found at the very start of it".

AlfnXd:
> What router model and browser are you seeing an issue? The original code looks good to me. indexOf() returns the position the string is found, or -1 if it's not found. Original code says "if the string is found anywhere within", yours simply says "if the string is found at the very start of it".

I am not using the asuswrt-merlin now, but I am using asuswrt 3.0.0.4.382_51129 on RT-AC1200G PLUS, however, for example, when `_cur_channel` is `36`, the condition `!= -1`  is satisfied for Both `"36/80"` and `"136/80"` in `wl1.channel_80m`,  which will cause `_wl_channel` to be pushed twice and the duplicate options in the Control Channel when user choose Channel bandwidth to **80 MHz**. 

image

Like this, for example：
image

I checked the code in chanspec.js and there is no much difference between them in the main business, so I am worried that this problem also exists. If there is no such problem, i will close this pull request.

AlfnXd:
@RMerl

PR #318, opened 2019-06-11
Fix AC5300 Network Tools file placement

dave14305:
This is more for the sake of experimenting with Github than really expecting it to be pulled.


RMerl:
I'm going to manually fix these rather than merge this, but a few recommendations for future pull requests:

1) They should be one single commits instead of two
2) It's best to reference what a commit is modifying.  In this case, the commit should be prefixed with "webui:" .

PR #330, opened 2019-06-30
Updated vpnrouting.sh for POSIX compliance

Xentrk:
Used https://www.shellcheck.net/ to update code for POSIX Compliance

### DIFF
--- a/release/src/router/others/vpnrouting.sh
+++ b/release/src/router/others/vpnrouting.sh
@@ -1,250 +1,214 @@
 #!/bin/sh
 
 PARAM=$*
-if [ "$PARAM" == "" ]
-then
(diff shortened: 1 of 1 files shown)


RMerl:
This patch doesn't match with my existing coding style, as it replaces tabs with spaces for indentation.

Xentrk:
> 
> 
> This patch doesn't match with my existing coding style, as it replaces tabs with spaces for indentation.

Okay.  What editor do you use?  

The reason for the spacing differences is I used the format shell script package in Atom editor.  I'll make another pass at it using NP++ if I have time.

Xentrk:
Updated code and created new pull request.

PR #342, opened 2019-08-03
remove the radio off confirmation introduced with aimesh

bughit:
every time you turn off the radio (which is already a multi-click operation), aimesh enabled firmware confirms with: 

> "If you disable Radio, it will affect the AiMesh wifi connectivity.\nAre you sure to process?"

This wasn't there before aimesh and it's not needed now. The merlin audience does not need to be reminded endlessly that turning off the wifi radio will affect wifi connectivity.

### DIFF
--- a/release/src/router/www/Advanced_WAdvanced_Content.asp
+++ b/release/src/router/www/Advanced_WAdvanced_Content.asp
@@ -761,11 +761,6 @@ function applyRule(){
 		}
 		
 		document.form.wl_sched.value = wifi_schedule_value;	
-		if(amesh_support && (isSwMode("rt") || isSwMode("ap"))) {
-			var radio_value = (document.form.wl_radio[0].checked) ? 1 : 0;
-			if(!AiMesh_confirm_msg("Wireless_Radio",radio_value))
-				return;
-		}
 		showLoading();
 		document.form.submit();
 	}


bughit:
@RMerl You don't want to merge this?

RMerl:
After reviewing the code, I believe keeping this message in place is important.  If you have nodes that are connected over wireless and you disable wifi, you will never be able to reconnect these nodes and will be forced to do a factory default reset of the nodes to be able to reconnect them.  Therefore, it's important that users be notified that changing this setting can have an important impact on the whole mesh network.

bughit:
> connected over wireless and you disable wifi, you will never be able to reconnect these nodes and will be forced to do a factory default reset of the nodes to be able to reconnect them

@RMerl 

That seems incredible. Are you saying this aimesh can't survive a power failure? Why wouldn't the nodes be able to reconnect when the master's radio is turned on?

RMerl:
No.  What I'm saying is if you make a change to the primary router wifi configuration (such as disabling wifi), that change will be pushed to the nodes, who will also disable their own wifi.  If that was the only way these nodes were able to connect to the primary router (if not using a wired backhaul), then you won't have any way to push the request to re-enable wifi to these nodes without resetting the nodes.

That's at least the theory that would explain why Asus explicitely added a warning - I cannot confirm the exact list of wifi parameters synchronized between the master and the nodes, as the cfg_sync source code is closed.

If Asus explicitly added the warning for AiMesh-enabled models, that's the rational reason I could see for it.  They have a number of such warnings handled through AiMesh_confirm_msg() calls.

bughit:
> that change will be pushed to the nodes, who will also disable their own wifi

I see.  Don't all aimesh routers have a wifi physical buttons? Also their warning is useless anyway as it does not explain what you just did.

But if you think it's needed, I'd like to suggest an alternative, being able to disable/hide aimesh, in the manner it was hidden/disabled by default in prior versions.  I am assuming this would disable aimesh specific alerts, like this one.

Those who don't need aimesh, which I would guess is a significant %, probably don't want any changes it brings, especially additional modal alerts, and would be happy to hide/disable it.

PR #416, opened 2020-01-30
rc: fix dnsfilter IPv6 additions to dnsmasq.conf

dave14305:
DNSFilter is only writing IPv6 options 5 & 6 to dnsmasq.conf since the other choices have an index 13 or greater. This change allows the remaining options to be written.

### DIFF
--- a/release/src/router/rc/dnsfilter.c
+++ b/release/src/router/rc/dnsfilter.c
@@ -256,7 +256,7 @@ void dnsfilter_setup_dnsmasq(FILE *fp) {
 
 	defmode = nvram_get_int("dnsfilter_mode");
 
-	for (dnsmode = 1; dnsmode < (sizeof server_table / sizeof server_table[0]); dnsmode++) {
+	for (dnsmode = 1; dnsmode < (sizeof(server_table)/sizeof(server_table[0])); dnsmode++) {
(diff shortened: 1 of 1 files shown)


RMerl:
To avoid the same issue from happening again in the future, the check should be made against the struct size rather than a hardcoded value, i.e.

```
for (dnsmode = 1; dnsmode < (sizeof server_table / sizeof server_table[0]); dnsmode++) {
```

dave14305:
I feel like I’ve now copied someone else’s homework! Thanks for the sound advice!

dave14305:
I’m kinda making this a sloppy mess, so don’t hesitate to do this yourself in one commit.

RMerl:
> feel like I’ve now copied someone else’s homework!

Nah, it's called "patch review", and is a normal part of handling pull requests/patch submissions.

I'll see if I can merge them into a single patch, otherwise I will just manually apply it.

asaddon:
"I feel like I’ve now copied someone else’s homework!"

That's the whole point of PRs, isn't it? Copying and merging other people's homework lol.

RMerl:
I had forgotten that these structs were local and not global.  I'll take the opportunity to also do some other planned changes while fixing this.

PR #433, opened 2020-02-22
httpd: Fix wireless password strength check with WPA3

Adamm00:


asaddon:
@RMerl already fixed it in https://github.com/RMerl/asuswrt-merlin.ng/commit/92c32ae843d0d2091a2ba28ca0a106d472a52431

Adamm00:
Different bug (but similar), this is for the password strength indicator, not the wireless encryption

asaddon:
Ah Okay!!

Adamm00:
Did a test compile just to make sure it works;

Before;
https://i.imgur.com/Dmu6hQU.png

After;
https://i.imgur.com/H36N4m4.png

PR #463, opened 2020-03-29
Update DoT

maghuro:
Changed cloudflare, as they're using one.one.one.one as main one.
Add Applied Privacy on port 443 which is useful for European users, because servers are based in Austria (country with hard privacy laws)


maghuro:
Forget about e9545c6, don't know why was it assumed as a pull request from me.

dave14305:
Cloudflare is still serving a certificate with CN=cloudflare-dns.com on 1.1.1.1:853.

maghuro:
You're absolutely right, my very bad. A better search and I read that.

maghuro:
@dave14305
I was testing and I can only use cloudflare's DoT using 1dot1dot1dot1.cloudflare-dns.com or one.one.one.one
Just cloudflare-dns.com it won't connect...

RMerl:
The official Cloudflare documentation still states cloudflare-dns.com should be used.

https://developers.cloudflare.com/1.1.1.1/dns-over-tls/

maghuro:
Correct
And regarding Applied Privacy?

RMerl:
No plan to add it.  I want to keep the list short, with either well-known services, or services that provide something unique (hence the presence of one service that uses a non-standard port).  Too many services will only confuse the users, and will increase my workload as I will have to constantly add or remove services.

PR #466, opened 2020-04-02
Make use of all icons present in client-list.svg

hitched97:
Make use of all icons present in client-list.svg
Fix some typos and trailing whitespace


RMerl:
Rejected.  Do not make formatting changes that are unnecessary, as this will make it very difficult for me to merge future GPL updates from Asus, since it will cause the diff/merge process to fail, and require me to start merging everything manually.

PR #529, opened 2020-05-29
unlock all languages for everyone

paldier:
BR CN CZ DE EN ES FR HU IT JP KR MS NL PL RU RO SL TH TR TW UK
you can choose what you like on any asus's router

Signed-off-by: ZhaoWei Xu <paldier@hotmail.com>

### DIFF
--- a/release/src/router/httpd/Makefile
+++ b/release/src/router/httpd/Makefile
@@ -107,6 +107,7 @@ OBJS += web.o common.o nvram_f.o
 OBJS += aspbw.o initial_web_hook.o
 OBJS += apps.o
 OBJS += sysinfo.o data_arrays.o
+OBJS += unlock.o
 
 ifeq ($(CONFIG_REALTEK),y)
 OBJS += web-realtek.o
--- a/release/src/router/httpd/ej.c
+++ b/release/src/router/httpd/ej.c
@@ -253,7 +253,7 @@ do_ej(char *path, FILE *stream)
 	// Load dictionary file
 
 	lang = nvram_safe_get("preferred_lang");
-	if(!check_lang_support(lang)){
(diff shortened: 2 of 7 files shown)


JackMerlin:
I support this pull, but I think Eric will not be merged into the code repository, because there is a reason why asus chose not to provide some languages on some national models. if we do not respect the decision made by asus, then asus may be closed in the future. so, that will make more difficult to maintain this project.

RMerl:
I'll give it more thought, but my first reaction would indeed be to not merge this, as this could create a conflict with Asus.  If they moved that code into closed source components, it`s because they did not want me to override them.

JackMerlin:
@paldier @RMerl 


> I'll give it more thought, but my first reaction would indeed be to not merge this, as this could create a conflict with Asus. If they moved that code into closed source components, it`s because they did not want me to override them.



If let me  to choose, I would rather sacrifice multi-languages support to choose openness. because that's why this project is so powerful.

BTW, I did lose other languages about a year ago. I cannot see French again. I can only use English. My model is from China. so this change may indeed help me, but I don't want to see asus close the source code, and I don't want to see this change merged into a future release.

GravuTrad:
Strange to can't have a real multilanguage system with asus. Really tiring.

RMerl:
Blame it on marketing.

In any case, I decided not to play a cat and mouse game with Asus, and won't be merging this change in.  Sorry.

GravuTrad:
For sure it's normal. How to merge it in our perso nal router?

JackMerlin:
> 
> 
> Blame it on marketing.
> 
> In any case, I decided not to play a cat and mouse game with Asus, and won't be merging this change in. Sorry.

Wise choice, I support you!

PR #567, opened 2020-07-24
webui: fix double username check

maghuro:
and also focus on which field is null


RMerl:
Already fixed internally.

PR #569, opened 2020-07-27
Update updown-client.sh

Xentrk:
1. Remove unnecessary echos
2. Standardize use of quotes for variables and echo statements to output files
3. Standardize use of ${instance} in code for readability
4. Remove unnecessary backslash escape character before '.' when setting file name vars at start 
5. Tune code for VPN_IP_LIST to collect the six nvram parms
6. Replace code in back ticks with $()
7. Tune code checking if parm has values or no values with -n and -z operators

### DIFF
--- a/release/src/router/others/updown-client.sh
+++ b/release/src/router/others/updown-client.sh
@@ -1,40 +1,40 @@
 #!/bin/sh
 
-instance=$(echo $dev | sed "s/tun1//")
-filedir="/etc/openvpn/client$instance"
+instance=$(echo "$dev" | sed "s/tun1//")
+filedir="/etc/openvpn/client${instance}"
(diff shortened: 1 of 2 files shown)


RMerl:
Please provide separate PRs.  updown-client.sh (and updown-server.sh) are going away, replacement handlers from libovpn are currently being tested.

I will probably keep vpnrouting.sh since the vast majority of what it does is run shell commands anyway, so there would be little benefit to move it to libovpn.  I haven't made my final decision yet about this one.

RMerl:
updown-*.sh scripts have been replaced with libovpn functions.

Xentrk:
New pull request submitted. Sorry for the delay.

PR #571, opened 2020-07-27
Update wan.c

godcong:
fix build error: the funtion was changed by the latest commit


RMerl:
The master branch is just a staging area for the mainline and ax branches, it`s not expected to be buildable.  I backported the new libovpn code simply to make staging easier, there are quite a few things that won't build properly off the master branch.

PR #588, opened 2020-08-24
CI Builds

stephengroat:
Let me know if this is something that would be good to implement. I can get the rest of the builds in, just want to know if it's worthwhile and need some help (src dirs, toolchains, build targets, etc.)

Here's an example of the outputs (artifacts): https://github.com/stephengroat/asuswrt-merlin.ng/actions/runs/221474756

Can obviously get all of the artifacts associated with releases easily as well

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,13 +19,7 @@ jobs:
       run: sudo dpkg --add-architecture i386
     - name: apt-get update
       run: sudo apt-get update
-    - name: apt-get instal
-      run: |
-      sudo apt-get install libncurses5 libncurses5-dev m4 bison gawk flex \
-      g++ gengetopt git gitk zlib1g-dev autoconf autopoint \
-      libtool shtool autogen mtd-utils intltool sharutils libstdc++5 texinfo \
-      dos2unix xsltproc u-boot-tools device-tree-compiler python qemu \
-      gperf liblzo2-dev uuid-dev build-essential lzma-dev liblzma-dev \
-      lzma binutils-dev patch cmake libglib2.0-dev gtk-doc-tools \
-      libc6-i386 lib32stdc++6 lib32z1 libelf1:i386 libncurses5:i386 libc6-dev-i386 \
-      bc libtool-bin gnulib xutils-dev libltdl-dev
+    - name: apt-get install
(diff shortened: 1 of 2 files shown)


RMerl:
Not something I consider really useful unfortunately.  Unlike typical software projects, this firmware project has to be compiled differently for 10 different router models.  A test build of just one model (for example the RT-AC68U) wouldn't properly test building for any of the other platforms, leading to a false sense of "hey, this code builds fine".  Also at times certain model will fail to build for extended period of time as I might not have received compatible GPLs for that particular model.

Appreciate the thought and the effort though.

stephengroat:
No worries. Before I close, just want to clarify something

> A test build of just one model (for example the RT-AC68U) wouldn't properly test building for any of the other platforms, leading to a false sense of "hey, this code builds fine"

You understand that these builds can be parallelized out to cover all models, right? I was just covering 2 models, but all models can easily be covered. I just need help determining which models go with which source dir, etc.

> Also at times certain model will fail to build for extended period of time as I might not have received compatible GPLs for that particular model.

Less can be done about this, but I can show you how to allow builds to fail, etc.

RMerl:
Parallel building of 10-ish models will take a pretty good amount of time (and I'm not sure how much resources the typical CI service providers would allow for free).  On my 6-thread VM running on an i7 7700K, the newer AX models take between 30 and 40 minutes to compile when I do only one at a time.  Building all firmwares in parallel takes me close to two hours (come on AMD, I could use Zen 3 right now.. )

Models are currently not just folder-specific, they are also branch-specific.  AX models can only be built from the ax branch, and until last week the other models could only be build from the mainline branch.  Master was just acting as a common staging ground for both branches.  I've started working on the merge of the newer 386 code base from Asus which will at least re-unify them together, and move things back to the master.  But this is another ball that needs to be juggled with when automating builds, I might someday have to do it again, and move the actual buildable code to separate branches once more if Asus forks out their own code bases.

In the end, it's just more trouble than it's worth it IMHO considering the chaotic workflow of this project.

stephengroat:
Github Actions supports up to 20 concurrent builds, 72 hrs of total workflow time (https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits). From my tests, builds would take ~45 minutes (give or take, there's some scheduling waiting).

No worries, i'll close it out

PR #602, opened 2020-09-09
Make VPN policy routing mode work without requiring webui client rules

mgziminsky:
Ensure default route for ovpnc* table always gets set and that killswitch always adds prohibit rule.

As it is now, if you want to setup a VPN client using custom IPSET routing or similar, you have to either do it completely manually, or define a dummy client in the webui in order for the routing table to be built properly. Unless I'm missing something, there is no reason not to always create the client routing table. 

Aside from the base logic fix, I also changed the condition used for adding the client routing table default route, since I believe it was an actual bug. If it was indeed correct before, I will change that condition back. The line in question is 199/195 `if [ -n "$route_net_gateway" ]` -> `if [ -n "$route_vpn_gateway" ]`

Tested on an RT-AC68U by mounting my version over `/usr/sbin/vpnrouting.sh`

### DIFF
--- a/release/src/router/others/vpnrouting.sh
+++ b/release/src/router/others/vpnrouting.sh
@@ -163,7 +163,8 @@ then
 	if [ "$VPN_FORCE" -eq 1 ] && [ "$VPN_REDIR" -ge 2 ]
 	then
 		/usr/bin/logger -t "openvpn-routing" "Tunnel down - VPN client access blocked"
-		ip route change prohibit default table $VPN_TBL
+		ip route del default table $VPN_TBL
+		ip route add prohibit default table $VPN_TBL
 		create_client_list
 	else
 		ip route flush table $VPN_TBL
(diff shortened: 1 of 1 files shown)


RMerl:
It's been years since I've designed this, so the reasons being the design decisions are quite fuzzy by now, but after reviewing the code, here are my thoughts:

- The reason for just changing the default gateway rather than deleting the whole table is probably so not to lose any existing static rule that may have been pushed by the server, and ensure we only affect the default gateway (for example, imagine a remote server at work pushing both static routes and a default gateway).  That code should be left as it is IMHO.

- The check for a non-empty clientlist was probably just to better optimize things, as there was no reason in setting this table up if there was no client using it.  There should be no harm in making the change regardless of whether the clientlist is empty or not.

- the route_net_gateway check does seem like a mistake, and should rather check for route_vpn_gateway.  It never caused any real issue because that code is mostly a safety net for misconfigured servers.   This could indeed be changed to check for route_vpn_gateway.

mgziminsky:
Thanks for the review. 

Regarding your first point, that makes sense, but my change is still only changing the default route, not clearing the whole table. Any other routes defined by this script, manually, or pushed by the server would still be there. Doing it this way just ensures that the route does actually exist, since calling `ip route change` won't create it if it's not present. Since that prohibit rule is the crux of the killswitch, I think it's especially important to ensure that it gets created successfully.

RMerl:
You're right, I misread that as a table flush rather than just a route deletion.

Ensuring there's always a (prohibited) default route would ensure the client be able to get out using another default route as it walks down the rest of the routing rules.

PR #633, opened 2020-11-08
Enable UDP support for busybox/netcat

uncle-fed:
Enables UDP support for netcat to be able to send router stats to other systems over UDP easily


RMerl:
Could you use Entware's netcat instead?  Since this change would add 2.5 KB to Busybox, and probably be used by maybe 3 or 4 persons total in the entire userbase, so I'm reluctant to apply it.

uncle-fed:
You're the boss :-)

PR #635, opened 2020-11-15
Add exfat to kernel 2.6

umerov1999:


RMerl:
I'd rather not merge this since the rest of the firmware is not exfat-aware anyway, making it useless for regular USB disk usage.

umerov1999:
> I'd rather not merge this since the rest of the firmware is not exfat-aware anyway, making it useless for regular USB disk usage.

[exfat_support.tar.gz](https://github.com/RMerl/asuswrt-merlin.ng/files/5558154/exfat_support.tar.gz)
This code add full support mount exfat, but hasn't exfat-utils (fsck) work fine in 2.6 kernel

PR #664, opened 2020-12-27
Fixed OpenVPN 'Initializing' typo in EN dictionary

Reeceeboii:
`Initializing` misspelled as `Initialinzing` in the EN dictionary's `vpn_openvpn_init` key


JackMerlin:
I think if this problem comes from upstream, you may need to submit to asus instead of here, because the more changes here, the more work RMerlin needs to merge with each GPL.

RMerl:
I don't make changes to the dict files unless they are actually code-breaking issues, for sake of future GPL merge simplicity with upstream.

ryandesign:
> you may need to submit to asus instead of here

How? Perhaps this information could be added to https://github.com/RMerl/asuswrt-merlin.ng/wiki/FAQ.

Issue #699, opened 2021-02-13
(GMT+10) Canberra, Melbourne, Sydney timezone (and perhaps others) actual bug


snaunton:
**Describe the bug**

nvram variable time_zone_x, therefore /etc/TZ is not set property for Sydney timezone. This is caused by a bug in the C code.

**Model**

RT-AC68U

**Firmware version**

386.1_2

**Client information**

**To Reproduce**

Set to (GMT+10) Canberra, Melbourne, Sydney timezone in Administration system.

**Expected behavior**

date command should return correct time.

```
# nvram show | grep ^time_zone
time_zone_dst=1
time_zone_x=UTC-10DST,M10.1.0/2,M4.1.0/3
time_zone_dstoff=M10.1.0/2,M4.1.0/3
time_zone=UTC-10DST_1
# cat /etc/TZ:
UTC-10DST,M10.1.0/2,M4.1.0/3

```
**Actual results**

date command is 1 hour behind during DST period.

```
# nvram show | grep ^time_zone
time_zone_dst=1
time_zone_x=UTC-10DST
time_zone_dstoff=M10.1.0/2,M4.1.0/3
time_zone=UTC-10DST_1
# cat /etc/TZ
UTC-10DST
```

**Will this error be reproduced on the official firmware?**

Have not tested

**Screenshots**

**Additional context**

This seems to be caused by a bug in void time_zone_x_mapping(void) in release/src/router/rc/common.c:

release/src/router/rc/common.c:

```
void time_zone_x_mapping(void)
...
	len = snprintf(tmpstr, sizeof(tmpstr), "%s", nvram_safe_get("time_zone"));
	/* replace . with : */
	while ((ptr=strchr(tmpstr, '.'))!=NULL) *ptr = ':';
	/* remove *_? */
	while ((ptr=strchr(tmpstr, '_'))!=NULL) *ptr = 0x0;

	/* check time_zone_dst for daylight saving */
	if (nvram_get_int("time_zone_dst"))
		len += sprintf(tmpstr + len, ",%s", nvram_safe_get("time_zone_dstoff"));
...
	if ((fp = fopen("/etc/TZ", "w")) != NULL) {
		fprintf(fp, "%s\n", tmpstr);
		fclose(fp);
	}
```

The following line from the above copies "UTC-10DST_1" to tmpstr:

`len = snprintf(tmpstr, sizeof(tmpstr), "%s", nvram_safe_get("time_zone"));`

[...]

PR #701, opened 2021-02-16
time_zone_x_mapping bug fix

snaunton:
Bug for some time zones where time_zone_dstoff would not be appended to /etc/TZ:

https://github.com/RMerl/asuswrt-merlin.ng/issues/699


snaunton:
Enhancement: initial-clock-set user script. Quite handy to have the option execute things when the clock is set accurately.

RMerl:
Can you please provide a PR that contains only the first commit?  The second one is unrelated to it.

snaunton:
Whoopsie. Reverted.

PR #706, opened 2021-02-23
Use interval for clock & boot time

jackyaz:
Replace timeout to fix delay/sync of clock


RMerl:
What issue does this address?  They both run the same code every 1000, except in one case the function re-establish a new delayed call, while the other does it automatically.

jackyaz:
The setInterval call can end up being delayed amd nothing strictly run every 1s. This is noticeable if you leave the System Log page open for a few minutes, the system clock as shown on the page will grow increasingly slow vs. actual time.

PR #719, opened 2021-03-12
workaround:fix UU Game Accelerator startup

paldier:


RMerl:
At this point I'd rather not merge this, as I have no way of testing it, and when I contacted Asus to see if they knew anything about built-in Asuswrt-Merlin support added by the plugin devs they couldn't answer me.

PR #759, opened 2021-06-26
fix nf_conntrack: expectation table full

paldier:
Enhance the stability of all connections (also for fullcone).
I have tested it, the fullcone always keeps nat1 and will not become nat3 anymore.


RMerl:
Setting nf_conntrack_expect_max here will conflict with the value already set by setup_conntrack() in rc/common.c.

PR #785, opened 2021-07-27
Layout fix

grogi:
Some pages in the UI have a very weird margin above tabs.

image

This pull request applies the same technique (negative margin-top) used in other pages in the interface to counter this effect.

### DIFF
--- a/release/src/router/www/Advanced_Firewall_Content.asp
+++ b/release/src/router/www/Advanced_Firewall_Content.asp
@@ -424,7 +424,7 @@ function updateDateTime(){
 	  	<div id="subMenu"></div>
 	</td>	
     <td valign="top">
-		<div id="tabMenu" class="submenuBlock"></div>
+		<div id="tabMenu" class="submenuBlock" style="margin-top: -120px"></div>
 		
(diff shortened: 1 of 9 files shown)


RMerl:
That looks more like a workaround than an actual fix to me, as this would apply the offset regardless of whether that banner is shown or not.  I'd rather see a cleaner fix from upstream (so far I have never been able to reproduce it myself).

jackyaz:
i think this is being caused by jQuery not being loaded on the pages. state.js contains a function to dynamically adjust the content margin-top but uses jQuery references to do so. on working pages, jQuery is declared. on non-working pages, no jQuery

grogi:
> i think this is being caused by jQuery not being loaded on the pages. state.js contains a function to dynamically adjust the content margin-top but uses jQuery references to do so. on working pages, jQuery is declared. on non-working pages, no jQuery

I think you might be right.

I didn't invent the 'margin-top: -120px' - I pulled it from the rendered HTML (as shown below). After your suggestion I checked the source for that page and indeed cannot find the reference to it in the source code. 

image

I'll close this PR, as it is not a good solution.

PR #789, opened 2021-08-11
webui: fix broken Wireless tab

dave14305:
If any wl*_wpa_psk variable has special characters than can break JSON parsing from httpApi.nvramGet, the Wireless tab on the Network Map page will appear broken and QR codes will not display.

Since ssid and wpa_psk variables are retrieved separately (and properly URI decoded) after the initial httpApi.nvramGet call, exclude the wpa_psk variables from that initial call.

### DIFF
--- a/release/src/router/www/device-map/router.asp
+++ b/release/src/router/www/device-map/router.asp
@@ -72,12 +72,12 @@ function getVariable(){
 	if(system.band2gSupport){
 		var _element = new Array();
 		if(isSwMode('re') && (concurrep_support || wlc_band == '0')){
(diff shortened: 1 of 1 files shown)


RMerl:
Can you provide me with a specific test case that triggers the issue?

dave14305:
`nvram set wl0_wpa_psk='badpass\(word'`

Actual impacted users have been very reluctant to disclose actual specific combinations of special characters causing the issue.

RMerl:
This is something that Asus already has fixed upstream between 42095 (what I'm using) and 43252 (which I checked).

PR #790, opened 2021-08-16
dsl-ax82u: Added cake support to QOS MultiWAN

gnuton:
This adds Cake to router buoilding QoS for MultiWAN.
Not sure if it's good to duplicate the code .An option would have been to create a a qos_cake.h/c files that would also be an option.
In any case this is the patch for now...
This has been tested on DSL-AX82U


RMerl:
Moving it to its own file might have made more sense indeed, however since so far Asus kept both tQoS and aQoS in the same files, I preferred to keep it this way as well.

PR #793, opened 2021-09-30
ISRG Root X1 cert

umerov1999:
Update lets encrypt root cert


quagsirus:
ISRG Root X1 is now included in `ca-bundle.crt` so we don't need to manually add, but `DST_Root_CA_X3.pem` still needs removing as it expired in September 2021

RMerl:
Was manually removed recently, closing this.

PR #794, opened 2021-12-17
kernel 2.6.36: restore ip6_tables set-return code

kjbracey:
The netfilters in the 2.6.22 and 2.6.36 kernels have an addition to
their CONNMARK support that lets the CONNMARK target act as a RETURN via
the --set-return option.

This is signalled by the connmark target function returning `XT_RETURN`.

However this is not standard Linux behaviour - target functions are not
supposed to jump - so to support it `ipt_do_table` and `ip6t_do_table`
need added code that checks for and handles `XT_RETURN` in the same way
as they would handle the standard RETURN target.

Kernel 2.6.22 had this code. The initial 2.6.36 kernel was missing the
code, and had the `return XT_RETURN` in `connmark_tg` disabled. Later,
the code was restored to `ipt_do_table` and `return XT_RETURN` was put
back. However, the code was never restored to `ip6t_do_table`.

By inspection, this means that traditional QoS will not work correctly
for IPv6 traffic with kernel 2.6.36 - it may even cause traffic loss.

The LTS fork actually has the `return XT_RETURN` removed, which means
QoS categorisation won't work for either IPv4 or IPv6, but would avoid
packet loss.

This patch adds the necessary code to `ip6t_do_table`, which should make
everything work as well as it ever did.

Systems using newer kernels abandoned this `set-return` and use standard
`RETURN` targets instead, so will have no problem.

A possible alternative would be to modify userspace to always use
standard `RETURN` targets. This might avoid any other issues arising
from this non-standard extension, but would increase CPU load on old
systems due to the extra table entries.

### DIFF
--- a/release/src-rt-6.x.4708/linux/linux-2.6.36/net/ipv6/netfilter/ip6_tables.c
+++ b/release/src-rt-6.x.4708/linux/linux-2.6.36/net/ipv6/netfilter/ip6_tables.c
@@ -431,6 +431,14 @@ ip6t_do_table(struct sk_buff *skb,
 		verdict = t->u.kernel.target->target(skb, &acpar);
 		if (verdict == IP6T_CONTINUE)
 			e = ip6t_next_entry(e);
+		else if (verdict == IP6T_RETURN) {		// added -- zzz
+			if (*stackptr == 0)
+				e = get_entry(table_base,
+				    private->underflow[hook]);
+			else
+				e = ip6t_next_entry(jumpstack[--*stackptr]);
+			continue;
+		}
 		else
 			/* Verdict */
 			break;


kjbracey:
This is the result of inspection while messing with QoS on an RT-N66U with kernel 2.6.22 on the LTS fork. So I don't have a 2.6.36-based device to test with myself.

But pretty certain anyone using such a device with traditional QoS and IPv6 will be having problems.

kjbracey:
Actually, this may be effectively dead code. It may be the case that all devices using this kernel are already avoiding using `--set-return`. (Because it's broken?)

In the LTS fork I can see the `manual_return` flag in `rc/qos.c` is set to 1 for the RT-AC56U and RT-AC68U, which I believe are the models using src-rt.6.x.4708, and hence kernel 2.6.38. And it's set to 1 in this repo for basically all models in a huge switch statement.

I was led to believe it was live code because of this reversion in the LTS fork: https://github.com/john9527/asuswrt-merlin/commit/61f275ae54af9b4fd17dcfa57609cdf1861bd186

But maybe it was only the other change in that commit that was significant.

If you can confirm this is dead code, this can be closed. It instead might be a Good Idea(TM) instead to actually remove the set-return support from the 2.6.38 kernel to make sure no-one inadvertently tries to use it.

kjbracey:
Closing this as I now believe no-one should be using the `--set-return` stuff, and making it "work" is probably a Bad Idea.

PR #795, opened 2021-12-18
github: remove the issue template

JackMerlin:
For any support and feedback, please post in the SNB forums: 
https://www.snbforums.com/

Changelog from 386.3_2:
Closed down the Issue tracker on Github, as 90%
of it was people asking for technical support,
or failing to use the supplied submission form.


JackMerlin:
I don't think we need it anymore.

PR #798, opened 2022-01-05
openvpn: add missing cipher-negotiation.rst

galak:
Missing cipher-negotiation.rst causes build failure when docs try to
build.  Add file from openvpn 2.5.5 release.

Signed-off-by: Kumar Gala <galak@kernel.org>

### DIFF
--- a/release/src/router/openvpn/doc/man-sections/cipher-negotiation.rst
+++ b/release/src/router/openvpn/doc/man-sections/cipher-negotiation.rst
@@ -1,0 +1,96 @@
+Data channel cipher negotiation
+===============================
+
+OpenVPN 2.4 and higher have the capability to negotiate the data cipher that
+is used to encrypt data packets. This section describes the mechanism in more detail and the
+different backwards compatibility mechanism with older server and clients.
+
+OpenVPN 2.5 and higher behaviour
+--------------------------------
+When both client and server are at least running OpenVPN 2.5, that the order of
+the ciphers of the server's ``--data-ciphers`` is used to pick the the data cipher.
+That means that the first cipher in that list that is also in the client's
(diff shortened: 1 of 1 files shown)


RMerl:
Odd as that file does not exist in the OpenVPN 2.5.5 archive that I downloaded and merged in, and building the firmware is also working fine for me.

galak:
Here's the reference in the code base:

https://github.com/RMerl/asuswrt-merlin.ng/blame/master/release/src/router/openvpn/doc/openvpn.8.rst#L78

Here's the github openvpn repo I grabbed the file from:

https://github.com/OpenVPN/openvpn/blob/v2.5.5/doc/man-sections/cipher-negotiation.rst

I see that the tarball from openvpn doesn't seem to have the file.

[ will do a rebuild and get the exact error I was seeing ]

RMerl:
I can merge your patch in, not a problem.  Maybe you might want to check first if there might be more files missing from the OpenVPN tarball to add to this PR?

galak:
Looks like @kjbracey might have run into the same issue.

It seems like my issue might be not having cleaned out between updating to a new SHA.  So curious if @kjbracey can reproduce on a clean build.  Need to try a fresh clean build again to verify.

kjbracey:
IIRC I got the build error after cleaning openvpn (`make openvpn-clean`) to sort out some previous build glitch. It shows up when doing autoconf, I think.

It may be that some intermediate generated file has been checked in, but this original source is missing, so it works from a clean-checked-out build, but not if you do a make clean.

RMerl:
There are a lot of issues caused by a global "make clean" ever since the merge of the HND SDK by Asus, to the point that I have long given up trying to sort these out.  Having an app-specific clean like "make openvpn-clean" work properly is definitely helpful however.

PR #800, opened 2022-01-05
Assorted TQoS fixes

kjbracey:
* Fix transfer limits - undefined behaviour was stopping minimums from working
* Allow default rule to work when a transfer limit rule no longer applies
* Basic IPv6 support
* Fix imq0 download classification - mark restoration and masking
* Correct overheads
* Implement PTM
* Restore GUI display for download classification

Fuller detail in individual commit messages.


RMerl:
One thing to note: imq0 is not used by Asuswrt/Asuswrt-Merlin.  It's largely legacy code that came in when Asuswrt got forked from Tomato.  It required kernel support, and enabling it caused struct changes to skbuf which broke any precompiled module (like the wireless drivers).

PR #801, opened 2022-01-05
Add missing openvpn doc file

kjbracey:
The absence of this cause a build hiccup for me.

### DIFF
--- a/release/src/router/openvpn/doc/man-sections/cipher-negotiation.rst
+++ b/release/src/router/openvpn/doc/man-sections/cipher-negotiation.rst
@@ -1,0 +1,97 @@
+Data channel cipher negotiation
+===============================
+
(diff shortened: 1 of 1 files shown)


galak:
Same thing in PR https://github.com/RMerl/asuswrt-merlin.ng/pull/798

galak:
I was seeing an issue, but when I try a clean build I think it goes away.. so not sure if you might have some style files from doing an update of the sources like I did.

Curious if you see the issue if you clean out your tree or at least the openvpn dir.

kjbracey:
Oh, missed your version :)

IIRC I got the build error after cleaning openvpn to sort out some previous build glitch. It shows up when doing autoconf, I think.

It may be that some intermediate generated file has been checked in, but this original source is missing.

The build is generally a bit flakey about whether it needs to autoconf/automake, if you ever do anything like stashing and unstashing...

PR #802, opened 2022-01-09
Improve memory displays

kjbracey:
Separate out cache memory from the memory displays, avoiding confusion.

In the router status, "free" is changed from being MemFree to being (MemFree + Cached - Shmem + SReclaimable + Buffers), so it includes probably-claimable memory.  This is equivalent to htop's cache + buffers + free. "Used" is then physical total minus that, equivalent to htop's used + shared, plus the reserved memory not counted in Linux's total.

In the Tools, existing fields are left as-is, but new "Used" and "Available" rows are added. "Used" is equivalent to htop's used + shared, so matches the router status display, except that we're not including reserved memory. "Available" is the kernel's MemAvailable value - a good high-level space available indication.

Follows discussion at https://www.snbforums.com/threads/ax88u-high-ram-usage-maybe.76650/

### DIFF
--- a/release/src/router/httpd/sysinfo.c
+++ b/release/src/router/httpd/sysinfo.c
@@ -108,6 +108,66 @@ void GetPhyStatus_rtk(int *states);
 #define SI_WL_QUERY_AUTHE 2
 #define SI_WL_QUERY_AUTHO 3
 
+static const char * const meminfo_name[MI_MAX] =
+{
+	[MI_MemTotal] = "MemTotal",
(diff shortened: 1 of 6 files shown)


kjbracey:
There are still people getting confused by this: https://www.snbforums.com/threads/fix-this-bug-pls-big-file-size-copy-and-the-memory-getting-full-ver-386-4.76947/

JackMerlin:
awesome

Eldenroot:
Is this included?

PR #803, opened 2022-01-09
Avoid 32-bit overflow in GUI stats

kjbracey:
Ensure output performs multiplications in float or uint64. Problem only likely to hit with large swap files at the moment.

Fixes problem reported at

https://www.snbforums.com/threads/displayed-memory-usage-error-in-webui.76204/page-2#post-734474


RMerl:
For future pull requests, please prefix the affected components in the commit message.  i.e.:

httpd: Avoid 32-bit overflow in GUI stats

This helps keep the commit log useful.

PR #808, opened 2022-02-15
Typos

wilhelmloof:
In Swedish the Swedish language is called "svenska".
https://en.wikipedia.org/wiki/Swedish_language


RMerl:
I don't make changes to language files because mainainting these would be a major project on its own, and it would become difficult to keep in sync with future GPL merges.

wilhelmloof:
Ok, I didn't know that this type of changes could cause future problems. But know you know, how to spell the Swedish language in Swedish :D

Does Asus have any public development of AsusWrt? In other words can I send my requests upstreams?

PR #809, opened 2022-03-10
Change the BUG_ON() condition in brcmnand_send_cmd() which checks for the interrupt status "controller ready" bit to a WARN_ON.

ikrasnopeev:
Issue is being discussed here:
https://www.snbforums.com/threads/sporadic-gt-axe11000-reboots-due-to-kernel-panic.77173/

I did Blame for the version of brcmnand.c in https://github.com/torvalds/linux repo and found following commit, done about 6 years ago:
https://github.com/torvalds/linux/commit/422485da15e7e9e6f71a5efd1aa2248595cb486d
which seems will help at least to not to reboot the unit right away in the middle of something critically important.

### DIFF
--- a/release/src-rt-5.02axhnd.675x/kernel/linux-4.1/drivers/mtd/nand/brcmnand/brcmnand.c
+++ b/release/src-rt-5.02axhnd.675x/kernel/linux-4.1/drivers/mtd/nand/brcmnand/brcmnand.c
@@ -1336,7 +1336,7 @@ static void brcmnand_send_cmd(struct brcmnand_host *host, int cmd)
 			return;
 	} else
 #endif
-	BUG_ON(!(intfc & INTFC_CTLR_READY));
+	WARN_ON(!(intfc & INTFC_CTLR_READY));
 
 	mb(); /* flush previous writes */
 	brcmnand_write_reg(ctrl, BRCMNAND_CMD_START,
--- a/release/src-rt-5.02p1axhnd.675x/kernel/linux-4.1/drivers/mtd/nand/brcmnand/brcmnand.c
+++ b/release/src-rt-5.02p1axhnd.675x/kernel/linux-4.1/drivers/mtd/nand/brcmnand/brcmnand.c
@@ -1336,7 +1336,7 @@ static void brcmnand_send_cmd(struct brcmnand_host *host, int cmd)
 			return;
 	} else
 #endif
-	BUG_ON(!(intfc & INTFC_CTLR_READY));
+	WARN_ON(!(intfc & INTFC_CTLR_READY));
 
 	mb(); /* flush previous writes */
 	brcmnand_write_reg(ctrl, BRCMNAND_CMD_START,


raven-au:
IIUC the AXE11000 uses the 4.1 kernel source tree, Is that correct?
A post on the snb forums topic refers to 4.19 source which is not right, while a later post refers to the 4.1 source tree.

If I'm right what's probably needed is:

commit 9d2ee0a60b8bd9bef2a0082c533736d6a7b39873
Author: Kamal Dasu <kdasu.kdev@gmail.com>
Date:   Fri Mar 3 16:16:53 2017 -0500

    mtd: nand: brcmnand: Check flash #WP pin status before nand erase/program
    
    On brcmnand controller v6.x and v7.x, the #WP pin is controlled through
    the NAND_WP bit in CS_SELECT register.
    
    The driver currently assumes that toggling the #WP pin is
    instantaneously enabling/disabling write-protection, but it actually
    takes some time to propagate the new state to the internal NAND chip
    logic. This behavior is sometime causing data corruptions when an
    erase/program operation is executed before write-protection has really
    been disabled.
    
    Fixes: 27c5b17cd1b1 ("mtd: nand: add NAND driver "library" for Broadcom STB NAND controller")
    Signed-off-by: Kamal Dasu <kdasu.kdev@gmail.com>
    Signed-off-by: Boris Brezillon <boris.brezillon@free-electrons.com>

which is in addition to 422485da15e7 (the subject of this pull request) and possibly:

commit 3c7c1e4594efd57b98ae6f7298f40cff4f4fb47b
Author: Kamal Dasu <kdasu.kdev@gmail.com>
Date:   Tue Jun 4 10:36:29 2019 -0400

    mtd: rawnand: brcmnand: Refactored code to introduce helper functions
    
    Refactored NAND ECC and CMD address configuration code to use helper
    functions.
    
    Signed-off-by: Kamal Dasu <kdasu.kdev@gmail.com>
    Signed-off-by: Miquel Raynal <miquel.raynal@bootlin.com>

Ian

raven-au:
I was rather thinking the changes could be added to the Merlin 4.1 kernel source.
There are a couple of reasons for this.

These changes are already included in upstream 4.19 (so, for example, the AXE16000 has them and it's been reported that the problem doesn't appear to be present with this device).
There doesn't appear to be much change to this code so bringing the 4.1 code up to date with the 4.19 code should be fairly straight forward.

We could give it a try.

I'm not sure how this works for Merlin builds but, for example, OpenWRT has a process where it adds patches to the source during the build so that the original source is retained and one can see the individual changes in the patch files. How are updates added for Merlin builds?

I can do the back port (there may be a little more to bring it up to date) but I'm not sure I'll be able to build to test them and I don't have an AXE11000 to use to test either.
Also I'm pressed for time but I have annual leave coming up so I could do this near the end of December if it's agreed to include it.
Also I'm not sure how to go about this, I'm pretty sure I can't force push an update to this pull request to update it ... perhaps I will need to fork the repo and create a new pull request?

Ian

RMerl:
I'm always very cautious when it comes to kernel merges particularly if they touch critical components such as the nand driver, as a bug there could potentially destroy a router.

I will still forward this patch suggestion to Asus to see what their engineers think.

raven-au:
> I'm always very cautious when it comes to kernel merges particularly if they touch critical components such as the nand driver, as a bug there could potentially destroy a router.

Yes, that is a concern.

I guess I'm a little less wary since I'm faced with this type of decision quite often and working on file systems (mainly VFS and pseudo file systems, proc, kernfs (aka. sysfs) I often get stuck with things that are outside of the file system area (the proc file system is notorious for this) but related that no-one else is willing to touch or just wants to hand over to someone else.

I had a quick look at the history of this code and I was incorrect, this nand code didn't land in the upstream kernel until 4.2. And I'm not clear on the devices used in each of the routers so it's hard to work out if there are likely to be side effects from changes such as this and to work out what really are dependencies.

> 
> I will still forward this patch suggestion to Asus to see what their engineers think.

That's certainly a good idea, however this is probably better addressed to Broadcom engineers, although asus should be sufficiently familiar with the code ... I guess my concern is they might not be willing to ask Broadcom ... 

It's actually a pretty serious problem.

Ian

RMerl:
Asus' engineers said that the root cause of the reported kernel panics were not the brcmnand driver (I remember this is something they discussed with Broadcom when it was initially reported).  Therefore this patch wouldn't help with that, and to avoid any potential side-effect it's best not to merge it in.

raven-au:
> Asus' engineers said that the root cause of the reported kernel panics were not the brcmnand driver (I remember this is something they discussed with Broadcom when it was initially reported). Therefore this patch wouldn't help with that, and to avoid any potential side-effect it's best not to merge it in.

Right, they would know best.
Don't get me wrong, above I wasn't suggesting it be merged regardless, ;)

Ian

RMerl:
No problem, I still appreciate you taking the time to research this.

I'm going to close this PR for now.

PR #810, opened 2022-03-12
webui: QoS Classification page: various cosmetic fixes

dave14305:
Proposing some cosmetic changes to the QoS Classification page.

### DIFF
--- a/release/src/router/www/QoS_Stats.asp
+++ b/release/src/router/www/QoS_Stats.asp
@@ -59,12 +59,6 @@ var qos_default = 0;
 
 if ("<% nvram_get("qos_enable"); %>" == 0) {	// QoS disabled
 	qos_type = -1;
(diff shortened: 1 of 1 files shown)


RMerl:
qos_default should be assigned in the previous **if** blocks rather than being re-assigned in this location.

dave14305:
> qos_default should be assigned in the previous **if** blocks rather than being re-assigned in this location.

I moved the `if` block down since I need to ensure bwdpi_app_rulelist is populated first. Let me know what you think about that. Thanks for the review.

RMerl:
Moving the following block:

```
if ("<% nvram_get("qos_enable"); %>" == 0) {	// QoS disabled
	qos_type = -1;
```
down means that if QoS is disabled, then type won't be set to -1, which can lead to undefined results as the following block that handles bwdpi strings on qos_type == 1.

PR #811, opened 2022-03-16
httpd: implement ICMP support for QoS Stats tracked connections

dave14305:
ICMP traffic is tracked by Adaptive QoS, but only a 3-letter protocol is allowed by the bwdpi function. Increase the possible proto field length from 3 to 7 (e.g. tcp, udp, icmp, unknown).

IPv4 will show the proto as `icmp`. IPv6 shows the proto as `unknown`, but can be replaced with `icmp` if sport and dport equal `65535`.

This has not been compile-tested.

Sample lines from `/proc/bw_cte_dump` for pings:
```
ipv4 icmp src=192.168.1.118 dst=1.1.1.1 sport=65535 dport=65535 index=1755 mark=1090076
ipv6 unknown src=26:01:00:00:00:00:00:00:00:00:00:00:00:00:00: dst=26:06:47:00:47:00:00:00:00:00:00:00:00:00:11: sport=65535 dport=65535 index=2020 mark=1090076
```

### DIFF
--- a/release/src/router/httpd/data_arrays.c
+++ b/release/src/router/httpd/data_arrays.c
@@ -1114,7 +1114,7 @@ int ej_bwdpi_conntrack(int eid, webs_t wp, int argc, char **argv_) {
 	char line[256];
 	FILE *fp;
 	static unsigned int count = 0;
-	char src_ip[64], dst_ip[64], prot[4];
+	char src_ip[64], dst_ip[64], prot[8];
 	int dport, sport;
 	int ret;
 	unsigned long mark;
@@ -1186,7 +1186,7 @@ int ej_bwdpi_conntrack(int eid, webs_t wp, int argc, char **argv_) {
 
 	while (fgets(line, sizeof(line), fp)) {
 		// ipv4 tcp src=192.168.10.156 dst=172.217.13.110 sport=8248 dport=443 index=8510 mark=3cd000f
-		if (sscanf(line, "ipv%c %3s src=%63s dst=%63s sport=%d dport=%d index=%*d mark=%lx",
+		if (sscanf(line, "ipv%c %7s src=%63s dst=%63s sport=%d dport=%d index=%*d mark=%lx",
 			          &ipversion, prot, src_ip, dst_ip, &sport, &dport, &mark) != 7 ) continue;
 
 		id = (mark & 0x3F0000)/0xFFFF;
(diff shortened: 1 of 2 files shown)


dave14305:
This doesn't compile because of some data type issues in my added code, and my limited understanding of C. I'll leave this as a draft for now in case @RMerl is interested in the concept but knows how to do it correctly.

dave14305:
```
data_arrays.c: In function ‘ej_bwdpi_conntrack’:
data_arrays.c:1210:16: warning: character constant too long for its type
    if (prot == 'unknown') && ((sport == 65535) && (dport == 65535))
                ^
data_arrays.c:1210:13: warning: comparison between pointer and integer
    if (prot == 'unknown') && ((sport == 65535) && (dport == 65535))
             ^
data_arrays.c:1210:30: error: expected identifier before ‘(’ token
    if (prot == 'unknown') && ((sport == 65535) && (dport == 65535))
                              ^
```

RMerl:
You are comparing a char pointer with another char pointer, rather than their content.  You need to use strcmp() when comparing strings.

dave14305:
I think I've gotten past the issues. I don't yet know if I will end up with a usable build or not since my build env is sub-par. Please see if this looks sane to you. I increased the length of proto from 4 to 8 so it can fit "unknown" + 1 char.

dave14305:
Compiled. built, flashed and tested on 386.6 alpha1 code.

image

vladut123:
> Compiled. built, flashed and tested on 386.6 alpha1 code.
> 
> image

Can you show me the main page?

dave14305:
> Can you show me the main page?

What do you mean by the main page? Network Map (index.asp)? I don't think it has changed.

vladut123:
> > Can you show me the main page?
> 
> What do you mean by the main page? Network Map (index.asp)? I don't think it has changed.

Yes (index.asp)

RMerl:
@dave14305 Showing a port number makes no sense here tho, the display shouldn't show any port for ICMP packets.

@vladut123 This PR has nothing to do with the main index page, it's strictly related to the QoS Classification page.

dave14305:
> Showing a port number makes no sense here tho, the display shouldn't show any port for ICMP packets.

Since it was in the bw_cte_dump, I didn't want to second-guess it, or get into complicated decisions about how a blank port would affect the Port filter fields in the webui. But I get your point.

I will give it some thought unless you already have ideas where to suppress it (C or ASP).

vladut123:
> @dave14305 Showing a port number makes no sense here tho, the display shouldn't show any port for ICMP packets.
> 
> @vladut123 This PR has nothing to do with the main index page, it's strictly related to the QoS Classification page.

One more question why asuswrt_2020 was last seen on snbforums in december 13 2021?

vladut123:
> @dave14305 Showing a port number makes no sense here tho, the display shouldn't show any port for ICMP packets.
> 
> @vladut123 This PR has nothing to do with the main index page, it's strictly related to the QoS Classification page.

One last question when will ASUSWRT_2020 be active on snbforums last time was december 13 2021

RMerl:
I would handle it at the webui level, to ensure it also gets sorted properly.

dave14305:
The ports are hidden, but they still remain 65535 for table sorting/filtering purposes.

image

RMerl:
Keeping 65535 might cause weird sort issues if there are valid tcp/udp entries with port 65535, which might get mixed with filtered ICMP entries.  I would suggest replacing 65535 with 0 or 65536 for these ICMP entries, to ensure they do get sorted together, without interfering with other port 65535 entries.

dave14305:
Rebased, mostly for practice.

PR #813, opened 2022-04-04
webui: Display Last 24 Hours Totals in KB when requested

dave14305:
Keep units consistent by respecting user dropdown choice. Previous behavior reduced KBs to MB, GB or TB in order to keep the displayed sum less than 10000. Also use comma separators to keep larger numbers more readable.


dave14305:
I need to account for the other versions under sysdep.

PR #815, opened 2022-05-22
ddns: implement cloudflare.com support

mohd-akram:


paldier:
cf api
https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record
`     -H "X-Auth-Key: c2547eb745079dac9320b638f5e225cf483cc5cfdda41"`
inadyn
`	"Authorization: Bearer %s\r\n"	\`
Inadyn doesn't support api token, it uses password only, so it doesn't work right now.

mohd-akram:
API token is different from API key. See [here](https://developers.cloudflare.com/api/tokens/create/) and [example](https://github.com/RMerl/asuswrt-merlin.ng/blob/master/release/src/router/inadyn/examples/cloudflare-ipv4-only.conf). I've tried it, it works.

mohd-akram:
Is this good to merge?

RMerl:
I currently don't plan on adding more DDNS services, as there are simply too many and that would require actively maintaining them whenever there's an API change.

mohd-akram:
I understand not wanting to add too many, but doesn't inadyn handle API changes?

RMerl:
inadyn can take weeks, if not months to adapt to API changes (the author often has to rely on outside contributors as he doesn't have accounts for every supported providers).  So rather than having to deal with the flood of support requests that would come from my users if a provider is broken for 3-6 months, I prefer to keep the list as small as possible (largely only Asus-supported providers, with a few specific exceptions).  I provide everything needed for end-users to manually use different providers if they want to, but they do so on their own then.

PR #819, opened 2022-06-18
Proof of concept for RT-AX86u - Custom IPv6 DNS server config for DHCP clients

mkem114:
Hi! Sorry, this is my first time contributing to this repository, so please excuse noob questions. I would love any feedback please.

Myself and it seems some others would like this feature built into the firmware:

- https://www.snbforums.com/threads/asuswrt-merlin-serving-ipv6-router-ip-instead-of-ipv6-dns-server-ip-f-w-384-19.67225/
- https://www.snbforums.com/threads/router-is-overwriting-my-preferred-ipv6-dns-address.70991/
- https://www.snbforums.com/threads/asuswrt-merlin-ipv6.74893/
- https://www.snbforums.com/threads/ipv6-dns-server-ips-broadcast.77456/

Re: this feature

- Should we support more than one DNS server? Should advertising the router's be a separate option?
- Is it possible to get this adopted upstream (I mean Asus)?
- I realise that the branch I'm working off is not head, I wanted to use stable for my personal router. Can I confirm the branch I should work off when doing this for real? What's the new folder for building for the RT-AX86U?
- Should I add to the system log?
- Is there another place I should make changes that I don't know about?

I've built and have this running on my RT-AX86u and have confirmed this works. I've found one bug to fix which is that the enabled/disabled is inverted for showing the custom input field.

Unrelated to this PR; I've found a bug with the DHCP DNS server config. The second input field doesn't have `autocorrect="off" autocapitalize="off"`.

Thanks!


dave14305:
It’s already there (albeit slightly broken) in the latest beta.

https://github.com/RMerl/asuswrt-merlin.ng/blob/89778ae3690fbfe38cdfa66f6699f8fa6cb31428/release/src/router/www/Advanced_DHCP_Content.asp#L991

RMerl:
Rejecting this, as Asus is already working on their own IPv6 DNS implementation as of recent GPLs.

PR #820, opened 2022-08-18
Add if statement for specific routers to allow compilation on other models.

RSDNTWK:
This may be temporary however this pr would allow other models to be able to compile the latest version without the "undefined reference error: wl_apply_akm_by_auth_mode". We'd thought we'd share this in case anyone wishes to try out. The latest revision has been tested on the RT-AC88U and works just fine.


RMerl:
Having it compile does not mean that it will work properly, as you would be mixing prebuilt binaries with GPL code that is 2-3 months apart.  For instance the latest code adds support for UNII-4 channels, but the closed source components are much older and wouldn't properly handle these, leading to wifi problems.  Similar issues common happen when mixing older AiMesh binary blobs with newer GPL code.

PR #821, opened 2022-08-20
firmware update: add unique identifier for statistics

JackMerlin:
Add merlin_fwupdate_uuid mvram variable to persist a UUID, UUID is randomly generated, not bound to hardware, and will be reset after factory reset.

Statistics sent to the server are sent through the HTTP user agent in the format: Model-FirmwareVersion-UUID. (ex: RT-AX68U-386.7_2-323fd9e3-7c70-45af-b0de-57caec678e54)

When the merlin_fwupdate_uuid variable is 0, it means that the user has explicitly opted out of statistics collection, will stop sending the unique identifier to the server, and revert to the default behavior of the previous firmware.

### DIFF
--- a/release/src/router/rom/webs_scripts/merlin_webs_update.sh
+++ b/release/src/router/rom/webs_scripts/merlin_webs_update.sh
@@ -1,6 +1,23 @@
 #!/bin/sh
 
-wget_options="-q -t 2 -T 30"
+if [ -z "$(nvram get merlin_fwupdate_uuid)" ]; then
+	# If there is no UUID then generate the UUID and save it as a persistent variable for later use.
+	# A factory reset will make it regenerate.
+	uuid="$(cat /proc/sys/kernel/random/uuid)" # generate a uuid
+	nvram set merlin_fwupdate_uuid="$uuid"
+	nvram commit
+fi
+
+if [ "$(nvram get merlin_fwupdate_uuid)" = "0" ]; then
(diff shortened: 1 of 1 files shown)


RMerl:
This is of no use since the update server is behind a CDN, and therefore the query never reaches the server itself.

JackMerlin:
> This is of no use since the update server is behind a CDN, and therefore the query never reaches the server itself.

Since this data is sent to the server via the user agent, the log records the user agent for you. If you're using Cloudflare, here's an introduction to their logging information:

https://developers.cloudflare.com/logs/reference/log-fields/zone/http_requests/

And you can even block check for update requests from illegal hardware via user agent
https://support.cloudflare.com/hc/en-us/articles/115001856951-Understanding-Cloudflare-User-Agent-Blocking

JackMerlin:
Related discussion: https://www.snbforums.com/threads/advantages-disadvantages-of-installation.80387/post-785027

JackMerlin:
This is just a proof-of-concept code as it is very simple to implement.

The UUID is used to determine the uniqueness of each update request, and the UUID will be reset after the router is factory reset.

Firmware and Model are used for statistics.

Since this information is sent through the user agent, more granular management can be done in the server admin panel or in Cloudflare's dashboard.

JackMerlin:
Closed because I've elaborated my idea through code.

Any further discussion should go to the SNB forum :)

PR #828, opened 2022-12-09
busybox:disable removal of critical system files

paldier:
busybox:disable removal of critical system files on bcm4912 or bcm6756


RMerl:
Asus has their own implementation of a list of filtered critical files, which does include these paths.

PR #834, opened 2023-05-01
rc:fix failed to format jffs for HND 5.04 models

paldier:
we have to ensure that umount jffs always succeeds, which usually doesn't succeed in asus code. 
now, we can rework JFFS format option for all of HND 5.04 models.

### DIFF
--- a/release/src/router/rc/init.c
+++ b/release/src/router/rc/init.c
@@ -2730,8 +2730,14 @@ static void shutdn(int rb)
 	config_jumbo_frame();
 #endif
 #ifdef RTCONFIG_HND_ROUTER_AX_6756
-	if (f_exists("/jffs/remove_hidden_flag"))
+#ifndef MNT_DETACH
+#define MNT_DETACH	0x00000002
+#endif
+	if (f_exists("/jffs/remove_hidden_flag")){
+		if (umount("/jffs"))
+			umount2("/jffs", MNT_DETACH);
 		mtd_erase_misc2();
+	}
 #endif
 	printf("\nStopping bcm_boot_launcher ...\n");
 	system("bcm_boot_launcher stop");


paldier:
````
	if (strcmp(script, "reboot") == 0 || strcmp(script,"rebootandrestore")==0) {
#if defined(RTCONFIG_HND_ROUTER_AX_6756)
		if(nvram_match("jffs2_format", "1")){
			nvram_set("jffs2_format", "0");
			eval("touch", "/jffs/remove_hidden_flag");
		}
#endif
````

**Comment by paldier** [comment]
````
int reboothalt_main(int argc, char *argv[])
{
	int reboot = (strstr(argv[0], "reboot") != NULL);
	int def_reset_wait = 20;
#if defined(RTCONFIG_HND_ROUTER_AX_6756)
	if(nvram_match("jffs2_format", "1")){
		nvram_set("jffs2_format", "0");
		eval("touch", "/jffs/remove_hidden_flag");
	}
#endif
	nvram_set("sys_reboot_reason", "rbt_manual");
````

**Comment by RMerl** [comment]
HND 5.04 does not use the JFFS filesystem however, it uses UBIFS (even tho the mount point still uses /jffs/).   More work is required to implement support for it.  I started doing some work back in the day on rc/ubifs.c, but I don't remember what else was missing when I stopped working on it (beside the code to handle erasing the partition, Asus' code only handles the filesystem creation).

**Comment by paldier** [comment]
jffs2_format->ubifs_format
and add
````
	if (!check_if_dir_exist("/jffs/scripts/")) mkdir("/jffs/scripts/", 0755);
	if (!check_if_dir_exist("/jffs/configs/")) mkdir("/jffs/configs/", 0755);
	if (!check_if_dir_exist("/jffs/addons/")) mkdir("/jffs/addons/", 0755);
	if (!check_if_dir_exist(UPLOAD_CERT_FOLDER)) mkdir(UPLOAD_CERT_FOLDER, 0600);
```
That's all.
You should ensure that ubifs_format always is 0 when start_jffs2() starts.

RMerl:
I am now considering removing the option for all models.

1) The option was largely required for the RT-N66U and RT-N16 because their partition would be resized when flashing new firmware, which could sometimes lead to corruption, where re-formating would help
2) It was still somewhat useful to other models as people were storing data, scripts or configs in it, so the setting was kept moving forward

Now, both Asus and myself are using it to store critical information such as numerous router settings.  Erasing the JFFS partition without erasing the rest of the configuration will lead to unexpected results, as you will lose parts of your router configuration, which might be confusing to users not being aware about it.  Therefore the most common use scenario would be when resetting to factory default, and for that Asus has already put a checkbox in place on the Reset page to also erase the JFFS partition at the same time.

So for consistency and to avoid users unknowingly erasing part of their router configuration, there`s a solid chance the setting will be completely removed in a future release.

sulaiman-coder (approved):

dave14305:
> So for consistency and to avoid users unknowingly erasing part of their router configuration, there`s a solid chance the setting will be completely removed in a future release.

Sounds like this PR can be closed after https://github.com/RMerl/asuswrt-merlin.ng/commit/b6ffe7cf253251126aef25aa1b5b8c24146a731c

PR #835, opened 2023-05-06
MiniDLNA hungarian translation

ViBE-HU:


RMerl:
I don't touch translation content since it's virtually impossible to keep up to date afterward.

ViBE-HU:
Not clearly understand you but MiniDLNA never updates. No need to update the translation file either in the future cause it's done. It just annoyed me that it has been made in an era when adding multilang support for a little tool like this was not a common thing and every router software just simply forked the application without sources. Now Merlin has the sources and the possibility make it more friendly for this little change for some users. Of course it is your choice. But since there is no way to manually add it is a bit confusing.

PR #837, opened 2023-12-22
Never delete nvram.nvm

paldier:
Make sure that the file should not be deleted under any circumstances, including the full path or just the filename or `*`, such as `rm -rf /tmp/*`.


RMerl:
Asus implemented something similar upstream in coreutils/rm.c instead (for a different file).  Wouldn't it be better to match their implementation, or is there an issue with theirs?

paldier:
You can try running `rm -rf /tmp/*` or  `rm -rf *` in /tmp folder,  you will get the answer.
I have explained to asus, why is `rm -rf /tmp/*` or  `rm -rf *`, asus don't seem to believe it.

PR #839, opened 2024-01-30
upgraded inadyn to v2.12.0

bmlong137:
I am not sure if I did this right, but didn't seem too complicated

This will add support to numerous DDNS providers, including a critical fix for CloudFlare API keys.  The move of Google Domains to Squarespace sparked this, because Squarespace doesn't have DDNS.  I figure there may be others in the same boat.


RMerl:
That PR would completely overwrite Asus's numerous code changes, which would break a lot of things.

Upgrading inadyn is currently very difficult due to the extent of those changes done by Asus.  That`s why I had to stop updating it.

bmlong137:
Would it be feasible to instead just commit changes to the plugins?  That way we can add support for new endpoints and changes to existing ones?

(as long as the new plugins compile under the old headers)

RMerl:
Doubt it, there was some lower level changes in inadyn, in part for the IPv6 support implemented by Brainslayer.

PR #845, opened 2024-03-18
Fixed sorting when clicking on table header.

CKret:
The table sorter is flawed.
When clicking a header it flips the sorting direction every time. Regardless if the header is already selected or not.
Expected behaviour:

- Default sorting direction is increasing on the first column. (Except for Site Survey where signal strength is sorted by descending order by default).
- Clicking a column header that is already selected: flip the direction.
- Clicking a column header which is **not** selected: the sorting direction defaults to increasing.


RMerl:
Thank you.  I also added commit 729448ec4e9436e9c89a9dc166f78466cbd69816 which will ensure that RSSI sort will also default to decreasing order.

PR #846, opened 2024-03-30
[Security] Preliminary OpenSSL 3.0 support.

RSDNTWK:
We've added initial OpenSSL 3.0 support wherever possible to replace OpenSSL 1.1.1 that ended support on September 11th 2023 and has vulnerabilities CVE-2023-5678 & CVE-2024-0727 since the last release that are no longer patched. It currently works on the RT-AC88U and may work on others. We made this attempt to address the obsolete version by ensuring almost all existing components run on OpenSSL 3.0.13 while still retaining legacy support for components that rely upon them to attempt to minimize the vulnerability risks.

This is mostly a test pull request which may need a lot of cleanup and refinement. Feel free to make changes as necessary.


RMerl:
Upgrading OpenSSL is something already on Asus' roadmap, so I'm leaving it to them instead of devoting all the efforts into doing only a partial implementation, and switch to their work just a few months later once they finalize it.

For now it would make more sense to just backport any potential security issue that apply to 1.1.1.  Distros like Ubuntu are still backporting patches to 1.1.1, so their work could be reused.

RMerl:
Another issue with the dual-OpenSSL approach is the number of components like rc and httpd that have closed source components that are pre-compiled.  You can run into weird issues if rc is linked against curl which is linked against openssl 3.0.0, while rc has precompiled objects that make function calls designed for OpenSSL 1.1.1 and may not be compatible with 3.0.0.

Any dual OpenSSL port would need to be limited to very specific components that wouldn't run into such issues, like OpenVPN.

PR #850, opened 2024-06-12
Fix filter_wl_default_x type.

Invincibl-e:


RMerl:
That code is fine.  If the default policy is DROP, then you want to ACCEPT specific clients, in which case logaccept is the proper target.  Only the final rule in the chain will need to be logdrop, that isn't done through ftype.

Invincibl-e:
> That code is fine. If the default policy is DROP, then you want to ACCEPT specific clients, in which case logaccept is the proper target. Only the final rule in the chain will need to be logdrop, that isn't done through ftype.

The nvram_match use strcmp, If fw_default_x is equal, then return false, bro.

PR #852, opened 2024-06-14
Improve UPnP IGD & PCP/NAT-PMP

Self-Hosting-Group:
- Improve wording to include `UPnP IGD & PCP/NAT-PMP` and follow upstream change https://github.com/miniupnp/miniupnp/commit/02da7055fcec0dfd2903ab1e3f7d050f3ffb90e0

- Change default `upnp_ssdp_interval` to 900s to follow specification and upstream change https://github.com/miniupnp/miniupnp/commit/9339f0e52c7a1d2024377e7a2446e80b93d334aa


RMerl:
This PR has been closed as merged, so any new commit would need to be in a new PR.

RMerl:
The 386.x branch is now on limited support.  Asus already put all of these models on their EOL list last spring, and I am only providing necessary bugfixes for this branch until the end of 2024.

RMerl:
I would rather not.  Most 386 models are still running on kernel 2.6.36 and an old uclibc, so things related to networking and iptables are more problematic than on newer models.  These models are less than 6 months away from retirement, so after this last GPL merge to bring them more up-to-date, I don't want to deal with anything beyond simple updates and fixes over the remaining months.

RMerl:
IGDv2 support was problematic in the past on these older platforms, due to either the 2.6.xx kernel or the use of an antique uclibc instead of glibc (different IPv6 stack, and also some missing libc functions in uclibc).  That`s why I`m leaving the old 386 devices on their current implementation.

PR #853, opened 2024-06-19
General improvements, resolve alpha/beta comparisons

ExtremeFiretop:
Issue originally reported for the RT-AX86U running: 388.6_alpha1-g4889cee45a
Here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/issues/147

# Test Environment:
1. Current production update detection script (webs_update.sh)
2. Current nvram version: 388.7_0 (or 388.7_0_rog in my case)
3. Set nvram extendno from: 0 (or 0_rog in my case) to any of the below examples.

- **Example/Test Extendno's (Not working):**
- `0_alpha1-g4889cee45a`
- `0_beta3`
- `0_Beta3`
- `Beta1`

4. Check for updates. NO UPDATES FOUND!!! When there should be.

# Next Test Environment:
1. Current production update detection script (webs_update.sh)
2. Set nvram extendno from: 0 (or 0_rog in my case) to any of the below examples.

**Example/Test Extendno (Working):**
- `beta1`
- `alpha1-g4889cee45a`

3. Check for updates. UPDATES FOUND!!! (As expected)

# Final Test Environment:
1. Patched update detection script with my version (webs_update.sh)
- `mount -o remount,rw /`
- `curl --retry 3 "https://raw.githubusercontent.com/ExtremeFiretop/asuswrt-merlin.ng-1/patch-1/release/src/router/rom/webs_scripts/merlin_webs_update.sh" -o "/usr/sbin/webs_update.sh" && chmod 755 "/usr/sbin/webs_update.sh"`

2. Set nvram extendno from: 0 (or 0_rog in my case) to any of the below examples.

**Example/Test Extendno's (NOW Working):**
- `0_alpha1-g4889cee45a`
- `0_beta3`
- `0_Beta3`
- `Beta1`

**Example/Test Extendno (Working):**
- `beta1`
- `alpha1-g4889cee45a`

3. Check for updates. UPDATES FOUND!!! (As expected). 
As you can see, this maintains the original functionality, but also extends support for the all the alpha releases, now all extendno examples are properly compared.


ExtremeFiretop:
@Martinski4Github for visibility

RMerl:
The extendno values that do not work are not supposed to work in the first place, as they are NOT valid values.  Value extendno values are:

alpha1 through alpha9
beta1 through beta9
0 through 99 (where usually even numbers = release, odd numbers = beta)

With potential git hash appended as -gHASHVALUE.  There is also a limit to the maximum length supported by AiMesh's config management but I do not recall what it is.

0_alpha1 is not a valid extendno value and would never appear in a firmware build.

ExtremeFiretop:
> The extendno values that do not work are not supposed to work in the first place, as they are NOT valid values. Value extendno values are:
> 
> alpha1 through alpha9 beta1 through beta9 0 through 99 (where usually even numbers = release, odd numbers = beta) Everything must be lowercase.
> 
> With potential git hash appended as -gHASHVALUE. There is also a limit to the maximum length supported by AiMesh's config management but I do not recall what it is.
> 
> 0_alpha1 is not a valid extendno value and would never appear in a firmware build. There is no point in the update script to support fictional values when the rest of the firmware will not support them either.

Replace 0, with 2. So for example, 388.6_2-alpha1-g4889cee45a

I'm only using 0 now, because I'm describing test scenarios that are valid now for replications, but this extends past this current firmware release.

If 388.6_2 is officially released. (Which it was) This is the official released extendno value:

And the user is on the alpha version of it, the same thing applies. The extendno would be 2_alpha1-g,etc

RMerl:
> Replace 0, with 2. So for example, 388.6_2_alpha1-g4889cee45a

Again, that is an invalid value.  Re-read the valid extendno values that I listed, 2_alpha1 is NOT a valid extendno.  It should be either **alpha1**, or **2** - not both.  A valid test/beta version after 388.6_2 would be 388.6_3.

ExtremeFiretop:
> > Replace 0, with 2. So for example, 388.6_2_alpha1-g4889cee45a
> 
> Again, that is an invalid value. Re-read the valid extendno values that I listed, 2_alpha1 is NOT a valid extendno. It should be either **alpha1**, or **2** - not both. A valid test/beta version after 388.6_2 would be 388.6_3.

I'm not talking about after 388.6_2. I'm talking about the alpha release version of 388.6_2

The extendno value would be 2. The alpha would be 2_alpha1.

The screenshot I took above is literally after flashing 388.6_2 so how can we say 2 is not a valid extendno?

What is the extendno of the alpha release of 388.6_2 then?

RMerl:
> What is the extendno of the alpha release of 388.6_2 then?
388.6_1.

Beta and alpha tags are ONLY used before the the final _0 release.  Any subsequent point release will use the even/odd numbering scheme I mentionned.

ExtremeFiretop:
> > What is the extendno of the alpha release of 388.6_2 then?
> 
> Just "1". For a full version number of 3004.388.6_1.
> 
> Beta and alpha tags are ONLY used before the the final _0 release. Any subsequent point release will use the even/odd numbering scheme I mentionned.

Well then i'm out of ideas as to what caused this users issue, other than a malformed manifest at the time for that model, since I couldn't recreate it with mine at the time. (Unless I added a capital in the extendno, or added a 0_ before the alpha.) Which was the point of this PR. If you don't think the improvements are necessary or worth your time to merge, then please close it.

Any chance you still have older alpha versions (not on your OneDrive) I can take away for testing then?

RMerl:
> Well then i'm out of ideas as to what caused this users issue, other than a malformed manifest at the time for that model, since I > couldn't recreate it with mine at the time. (Unless I added a capital in the extendno, or added a 0_ before the alpha.)

You would need to know the actual value he had in his extendno variable, the actual model of his router, and also to ensure that the remote manifest did contain an update, and that he wasn't blocking access to the update server.

> Any chance you still have older alpha versions (not on your OneDrive) I can take away for testing then?

They are just build snapshots that I may generate now and then and that I directly copy to the Onedrive folder, I don't archive them, sorry.

PR #854, opened 2024-06-24
webui: Fix PCP port maps not shown on IPv6 status page if IGDv2 disabled

Self-Hosting-Group:
add `Time left`, remove `Description` and clearer/more consistent table headers, add link to enable service and improve wording.

As UPnP IGDv2 IPv6 port maps (pinhole) do not support a description, and the PCP description option is not supported, the information in the field is unlikely to be useful, but the Time left field is. Calculating time left from timestamp to have same format like as on the IPv4 port forwarding status page.

screenshot2
screenshot1


release/src/router/www/Main_IPTStatus_Content.asp
```diff
@@ -31,42 +31,47 @@ function show_upnp() {
...
-			code += '</tr>';
+		} else {
+			code += '<tr><td colspan="6"><span>No active port forwards.</span></td></tr>';
 		}
-	} else {
-		code += '<tr><td colspan="6"><span>No active port forwards.</span></td></tr>';
 	}
 
 	code += '</tr></table>';
+	code += 'IPv6 port forwards are listed on the <a href="Main_IPV6Status_Content.asp" style="text-decoration: underline;">IPv6</a> status page.';
```
RMerl, line 74: IPv6 pinholes are not port forwards.  They are firewall rules that open the port, there is no forwarding done at the firewall level.

release/src/router/www/Main_IPV6Status_Content.asp
```diff
@@ -132,36 +132,43 @@ function show_pinholes() {
...
+	code += '<th width="32%">Remote IP</th>';
+	code += '<th width="9%">Rem. Port</th>';
+	code += '<th width="32%">Local IP</th>';
+	code += '<th width="9%">Local Port</th>';
+	code += '<th width="12%">Time left</th>';
 	code += '</tr>';
 
-	if ("<% nvram_get("upnp_pinhole_enable"); %>" == "0") {
-		code += '<tr><td colspan="5">Pinhole support is currently disabled.</td></tr>';
+	if ("<% nvram_get("wan_upnp_enable"); %>" == "0") {
```
RMerl, line 144: Checking for UPnP being enabled is not enough, upnp_pinhole_enable must also be set to 1 for pinholes to be enabled.  Both vars need to be checked.

release/src/router/www/Main_IPV6Status_Content.asp
```diff
@@ -132,36 +132,43 @@ function show_pinholes() {
...
 				code += '</tr>';
 			}
 		} else {
-			code += '<tr><td colspan="5"><span>No pinholes.</span></td></tr>';
+			code += '<tr><td colspan="6"><span>No active port forwards.</span></td></tr>';
 		}
 	}
 
 	code += '</tr></table>';
+	code += 'IPv4 port forwards are listed on the <a href="Main_IPTStatus_Content.asp" style="text-decoration: underline;">Port Forwarding</a> status page.';
```
RMerl, line 171: There is no need for this note, this page is for IPv6 information, nobody would go there and expect to see IPv4 port forwards to be shown.

release/src/router/www/Main_IPTStatus_Content.asp
RMerl: That would be more accurate indeed.

release/src/router/www/Main_IPV6Status_Content.asp
RMerl: It shouldn't.  The point of this setting is to specifically prevent clients from opening IPv6 pinhole in the firewall (hence its name), in addition to increasing backward compatibility.  The webui description also explicitely calls it "[Enable UPnP IGDv2 (IPv6 support)]".

RMerl:
I've reworked the patch and merged the changes with e2aa8af62fe7723596e47402e522fd4ab31e10ee focussing on the display changes such as semantics, and replacing the unused desc field with the time left field.

RMerl:
Added commit 20d86d4a362817e24ef3c777be875a8fe0244b3a to properly disable IPv6 support when done so on the webui.

PR #855, opened 2024-07-03
Fixed comparisons of F/W versions

Martinski4GitHub:
Fixed 2 bugs in the code that were incorrectly comparing versions between the currently installed F/W with the next F/W update available.


Martinski4GitHub:
Note that technically, the 2 bugs (same bug twice) have no effect on the current functionality because the 1st & 2nd comparison expressions are coded correctly so with the 3rd & 4th comparison expressions (where the wrong variable is being used for "`$firmver`"), the result should still be OK because there should be no case where "`$current_firm" -gt "$firmver`"

Nevertheless, the wrong variable is being used in the 3rd & 4th comparison expressions.

PR #856, opened 2024-07-05
Update merlin_webs_update.sh - Certificate Failure

ExtremeFiretop:
Update merlin_webs_update.sh - Certificate Failure

Due to changes to the website: https://fwupdate.asuswrt-merlin.net/manifest2.txt
This script no longer functions correctly, and fails with certificate errors.

image

**-TEST ENVIRONMENT:**

Downgrade to 388.6 (or change the nvram values)
Run the **webs_update.sh** script (/usr/sbin/webs_update.sh)
Confirm that no updates are detected by checking the values:

`webs_state_flag`
and
`webs_state_info`

Also confirm that the wlan_update.txt located here: /usr/tmp/wlan_update.txt is empty:

image

**-TROUBLESHOOTING**

Modify the the **webs_update.sh** script (/usr/sbin/webs_update.sh) to include set -u **(temporarily)**
And removed the -q from the `wget_options="-q -t 2 -T 30"` so we can see the errors.

Identified the certificate error here:

image

And confirmed that adding a bypass for the certificate worked and populated the /usr/tmp/wlan_update.txt as expected:

image

And that the values for 

`webs_state_flag`
and
`webs_state_info`

Are now correctly set.

image

### DIFF
--- a/release/src/router/rom/webs_scripts/merlin_webs_update.sh
+++ b/release/src/router/rom/webs_scripts/merlin_webs_update.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-wget_options="-q -t 2 -T 30"
+wget_options="--no-check-certificate -q -t 2 -T 30"
 
 fwsite="https://fwupdate.asuswrt-merlin.net"
 


ExtremeFiretop:
@Martinski4GitHub 

For visibility

RMerl:
The problem is your clock, not the update server nor the update script.  As per your screenshot, your clock was set to May 2018, hence the "Certificate not valid yet" error.

ExtremeFiretop:
> The problem is your clock, not the update server nor the update script. As per your screenshot, your clock was set to May 2018, hence the "Certificate not valid yet" error.

No way to adjust the client on the node.
The Primary router clock is fine:

image

RMerl:
I am NOT disabling certificate validation, this is an important security measure.

Nothing has changed upstream, the Cloudflare SSL certificate has never been valid in 2018.

RMerl:
The certificate validation protects against potential redirection or hijacking of the trafic to a malicious server pretending to be the update server.  Without it, anyone could redirect fwupdate.asuswrt-merlin.net just by modifying the hosts file or poisoning your local DNS resolution to point to a different IP address, where any web server running a self-signed certificate would serve you potential malicious content.

ExtremeFiretop:
> The certificate validation protects against potential redirection or hijacking of the trafic to a malicious server pretending to be the update server. Without it, anyone could redirect fwupdate.asuswrt-merlin.net just by modifying the hosts file or poisoning your local DNS resolution to point to a different IP address, where any web server running a self-signed certificate would serve you potential malicious content.

Understand, I just wanted you to understand where I was coming from when I identified this behavior. 

This is your project and your rules.
I won't fight your decisions. I closed the PR and set the clock myself. 

But I would expect a node online and communicating with the Internet and devices to sync it's own clock.

The reason for this PR was to workaround that, that is all. I wasn't trying to fight you on anything, these are only suggestions *NOT* ;) demands.

ExtremeFiretop:
@RMerl 

And to add to this, this was beta 388.8 on the node, so I realize beta means beta. Maybe that had a role to play? Am I correct that this something that should sync itself? Or am I completely off base?

RMerl:
Is that an AiMesh node or an AP?  If it's AiMesh, what are the router model for both the parent router and the node?

I don't remember if NTP is enabled/synced on either AiMesh nodes or APs.

ExtremeFiretop:
> Is that an AiMesh node or an AP? If it's AiMesh, what are the router model for both the parent router and the node?
> 
> I don't remember if NTP is enabled/synced on either AiMesh nodes or APs.

This would be a AiMesh setup.
Parent would be a GT-BE98 Pro running the latest stable.

And the node would be a GT-AXE11000 running beta1 of 388.8

Keep in mind AiMesh and syncing between these firmwares has been problematic outside of this. That's why I totally understand if your pushing this back towards the clock not syncing.

For example. If I set the parent to use "6GHz first" for the Wireless Backhaul. It still uses 5GHz with the AXE11000.

RMerl:
I don't have time to test this specific setup at the moment, but on my test AiMesh setup (an RT-AX92U node connected to an RT-AX86U_PRO), the clock on my node does properly sync itself at boot time.  Check your system log on the node.

```
May  5 01:05:44 ntp: start NTP update
...
Jul  5 12:27:41 crond[1371]: time disparity of 3245002 minutes detected
```
Most likely cause is your node failed to complete name resolution for the NTP server.

In any case, the issue is with your node clock, not with the update script.

ExtremeFiretop:
> In any case, the issue is with your node clock, not with the update script.

Already closed 😁

ab151670:
> app  @**_intalll.webs_state_[Zip=}_**
>Use=https://fwupdate.asuswrt-merlin.net/manifest2.txt

PR #857, opened 2024-07-25
httpd: add two signedness check in httpd/http.c

LinRaymond2006:
Two signedness checks are added to `wget()` function in `httpd/http.c`.
Without the checks, a wget request from the implementation to a malicious server with a crafted `Content-Length` parameter (or a valid `Content-Length` plus a crafted `Transfer-Encoding: chunked`) may lead to a consequent buffer over-read in the statement `len = fread(buf, 1, len, fp);`.


RMerl:
Thanks.  I'll merge it just to be safe, but do note that this function isn't used by the firmware (it's probably legacy code), so the router was never vulnerable to it.

PR #860, opened 2024-10-20
Fix namecheap DDNS with let's encrypt

bugproduction:
Use the hostname field with FQDN, but only store the subdomain - if it exists - to the inadyn.conf file. This allows DDNS with namecheap with let's encrypt. Now one has to decide one or another.


0794-8819-0610-5047 (approved):

0794-8819-0610-5047 (changes requested):
_[]()https://github.com/RMerl/asuswrt-merlin.ng/blob/namecheap/release%2Fsrc%2Frouter%2Frc%2Fservices.c

PR #862, opened 2024-12-20
UPnP IGD & PCP/NAT-PMP: Improvements

Self-Hosting-Group:
- Minor wording improvements
- Set `upnp_clean=0` NVRAM by default, as it is for cleaning up port forwards/maps without traffic, but also cleans up legitimate unexpired ones after a period (600s by default) and does not work since longer (upstream bug exists), it is also not defined by IGD standards. Expired port map cleanup functionality works correctly
- Invert/reword secure mode option to `Allow third-party port forwarding` and apply to PCP as well
- Rearrange table headers in port map listings with more client-centric view


RMerl:
It`s initialized as the same time as all other IPv6 chains:

https://github.com/RMerl/asuswrt-merlin.ng/blob/master/release/src/router/rc/firewall.c#L3525

RMerl:
Look at the code I pointed you to - the chain is called UPNP, not MINIUPNPD.

RMerl:
The chain names are completely configurable within miniupnpd.

RMerl:
It was working when implemented, and it still does.

```
admin@stargate:/tmp/home/root# ip6tables -L UPNP -vn
Chain UPNP (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    0     0 ACCEPT     tcp      ppp0   *       ::/0                 2606:xxxxxxxxxx/128  tcp dpt:4445
```

RMerl:
Don't try to compare Tomato and Asuswrt-Merlin's implementation too closely.  I wrote the Asuswrt-Merlin PCP implementation (nearly 10 years ago so I don't recall the details).  We use different chain names because I stuck to the chain names picked by Asus (which were probably derived from their original pre-Asuswrt firmware).  I probably changed the name of the IPv6 filter table back then to be in sync with the IPv4 chains.

> It might be better to use the latest upstream code... There have been a lot of configuration/build changes recently.

Not a priority - the current implementation works fine, and Asus has done a significant amount of customization over the years, making any update involve a non-trivial amount of work to harmonize patches.

Beside, the last update from upstream only dates back to March 2024, and since then very little of importance has changed.

RMerl:
> * Can I see what patches have been applied?

No.  There's no individual patches, the code was just modified through years of code changes specific to Asuswrt.  You'd have to diff it to determine what was changed over the years.

> * And very interesting, what is the latest miniupnpd version you saw in the latest ASUS sources and is IGDv2 support configurable there?

2.2.0.  They currently have no intention to implement IGDv2 support.

RMerl:
> All source code downloads from ASUS where dated where is one of the most recent to download. Never found a way.

They're on their support site, tho some models are currently missing the GPL tarball.  Here's one of the recent ones:

https://dlcdnets.asus.com/pub/ASUS/wireless/GT-AX6000/GPL_GT_AX6000_300610234793.zip?model=GT-AX6000

> But is miniupnpd compiled with IPv6 support, so that PCP and IPv6 will work there (ASUS factory firmware)?

They don't enable IPv6 support, and they don't create any of the IPv6 required chains either.

RMerl:
> Thanks for all your hints, they helped a lot. But currently there is only one PCP implementation, which is also used by FreshTomato/AsusWrt, right?

I don't know, I have never looked at Tomato's UPNP implementation.

> What do you think about this PR?

No thoughts at this time, I haven't reviewed it yet.

RMerl:
Wording commit: unlikely that I'll merge it.

Disabling upnp_clean by default: that would change the behaviour that has been in place for 15 years within Asuswrt.  Do you have an explicit issue that would support changing the already established behaviour?

Secure mode options: extending it to PCP does make sense, I can apply that.  However I don't like the wording, "third party" could mean anything from an end user's point of view, such as "anyone outside of my network".  I prefer to explicitely state that it relates to preventing a client from forwarding to a different IP, which is the wording users have been familiar with for many years now.

Rearranging table headers: I think either layouts are fine.  The current one is designed with the logic that the typical user wants to see what's the open port (so, the external port), and to what local IP that port is forwarded, which would be within the first three columns, and is fine as it is.  I see no real reason to change it.

RMerl:
Closing this because this PR is no longer valid - I had to do extensive history rewrites over the past week to remove sensitive content from old commits, and I have now pushed it to Github, invalidating any previous PR (closed and open).  You will unfortunately need to reclone the cleaned repo and re-create a fresh PR.

RMerl:
Wait a few days before creating any new PR.  Since all previous PRs are now invalid, I asked Github support to remove them.  I don't know if they will only remove affectec PRs, or all of them.  If you look at the history of this PR, it is no longer usable due to the changed hashes.

RMerl:
I will just manually apply the services.c changes since this is an actual fix, and drop the other changes.

PR #865, opened 2025-02-11
Porkbun API Hostname Update

sargunv:
https://porkbun.com/api/json/v3/documentation#apiHost

> Please note that porkbun.com is no longer supported as the hostname for our API. Please use api.porkbun.com as the correct hostname moving forward.


RMerl:
The built-in acme.sh was customized by Asus and is only used for their closed source libletsencrypt.sh, so I prefer not to make any additional changes to it.  That's also why I don't update it.

PR #866, opened 2025-02-12
Fix gcc-15 build

stevefalco:
gcc-15 complains that main() returns an int.

Without this, gcc errors out early in the build process.  I'm using Fedora 41 to run the builds.

### DIFF
--- a/release/src/router/config/Makefile
+++ b/release/src/router/config/Makefile
@@ -101,7 +101,7 @@ endif
 .PHONY: ncurses
 
 ncurses:
-	@echo "main() {}" > lxtemp.c
+	@echo "int main() {}" > lxtemp.c
 	@if $(HOSTCC) lxtemp.c $(LIBS) ; then \
 		rm -f lxtemp.c a.out; \
 	else \


stevefalco:
Ideally this would be added to all active branches, and not just master.

RMerl:
I guess the rest of the build is fine because most of it is compiled by the toolchain?

stevefalco:
Correct.  I've been able to get a clean build for the RT-AX88U Pro that I just bought.  I haven't tried it yet, but the zip file that the build produced looks reasonable.

RMerl:
I had to revert that commit.  Asus already fixed it with GPL 388_25373, already present on the 3004.388 branch - it showed up as a conflict when I tried to merge master on top of 3004.388.

stevefalco:
Good.  It looks like your 3006 branch still needs it, but perhaps Asus has fixed that one too and it just hasn't made it into your tree yet.

0794-8819-0610-5047:
> Good.  It looks like your 3006 branch still needs it, but perhaps Asus has fixed that one too and it just hasn't made it into your tree yet.

RMerl:
> Good. It looks like your 3006 branch still needs it, but perhaps Asus has fixed that one too and it just hasn't made it into your tree yet.

I intend to wait until the next 3006.102 release, which is probably quite a few months away.  If at that time I don`t have a GPL with the fix in place, I will probably backport Asus' own fix from the 3004.388 branch.

PR #867, opened 2025-02-14
Added CVE-2024-13176 fix for openssl 1.1.1w Based on OpenSSL 3.0 patch.

RSDNTWK:
Added CVE-2024-13176 fix for openssl 1.1.1w based on the following openssl 3.0 commit fixing the affected code.
https://github.com/openssl/openssl/commit/07272b05b04836a762b4baa874958af51d513844

Marking as Draft as currently undergoing testing.

Confirmed Working on RT-BE88U so far.

PR #868, opened 2025-02-22
Ported CVE-2024-13176 fix for OpenSSL 1.1.1w from Ubuntu Repository.

RSDNTWK:
The following OpenSSL-1.1.1w patch has been ported from Ubuntu Repository https://packages.ubuntu.com/focal/openssl

CVE-2024-13176 Fix timing side-channel in ECDSA signature computation.

Rework of #867


RMerl:
Thanks.  The chances of this being exploitable on Asuswrt are next to nil (as it requires either physical access, or something with lower latency than an Internet connection), but probably still best to be applied if only for code consistency.

PR #871, opened 2025-03-27
webui: set correct tab when switching to VLAN Profile sub-tab

dave14305:
state.js does not account for the presence of the DNS Director tab in this hard-coded tab count, leading to the "Switch Control" tab being highlighted when a user clicks the Profile sub-tab in the VLAN page.


RMerl:
I don't have that issue on my RT-AX86U_PRO.  What model did you test?

PR #878, opened 2025-06-24
webui: Update Control D DoT labels and IP addresses

dave14305:
Rename the dropdown menu items for Control D DoT to align with the free DNS choices offered on the controld.com website and update the DoT IPs to match the guidelines given at the link below:

https://docs.controld.com/docs/control-d-ip-ranges#dns

The chosen free service will depend on the tls auth name given.

Or just rip them all out since the full list is quite unwieldy.

### DIFF
--- a/release/src/router/www/dot-servers.json
+++ b/release/src/router/www/dot-servers.json
@@ -141,45 +141,45 @@
 			"dnspriv_spkipin" : ""
 		},
 		{
-			"dnspriv_label" : "Control D 1 (Family Filter)",
-			"dnspriv_server" : "76.76.2.4",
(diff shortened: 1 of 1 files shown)


dave14305:
The TLS Auth comments were read on the Linux stubby section.

https://docs.controld.com/docs/linux-platform

RMerl:
Gotcha.  The fact they seem to hide it behind a web configurator makes me suspect these IPs/names may change at any time.  Let's try going with these updated entries, and if it changes again in the future then I will consider just dropping them.  They do offer an Asuswrt-Merlin configuration command, probably mostly for their premium subscribers I suspect.

PR #887, opened 2026-04-12
to be deleted

bottobot:
To be deleted


RMerl:
Do not publish unpatched security fixes on a public repo.
Do not bundle multiple patches into a single giant patch.
Do not publish AI-written patches that you haven't personally tested/validated through actual test cases.

RMerl:
I would appreciate if you could delete the branch indeed, just to be on the safe side.  Thank you.

No need to resubmit.  You're the second person in less than a week to send me statical analysis from Claude.  A lot of the result targets code that isn't even used by Asuswrt-Merlin, or things that were already fixed upstream by Asus.  A lot of the reported lack of validation is also already taken care internally, either within the closed source components, or they require root access to be able to exploit, which basically means you already have full control of the device anyway.

PR #894, opened 2026-07-13
rc: set ddns IPv6 variable on successful custom DDNS update

dave14305:
ddns_custom_updated does not set the IPv6 flag in nvram. Set the flag if the IPv6 DDNS feature is enabled.

### DIFF
--- a/release/src/router/rc/services.c
+++ b/release/src/router/rc/services.c
@@ -24738,6 +24738,10 @@ ddns_custom_updated_main(int argc, char *argv[])
 	if ((argc == 2 && !strcmp(argv[1], "1")) || (argc == 1)) {
 		nvram_set("ddns_status", "1");
 		nvram_set("ddns_updated", "1");
+#ifdef RTCONFIG_IPV6
+		if (nvram_get_int("ddns_ipv6_update"))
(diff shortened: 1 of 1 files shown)


dave14305:
My C is weak and I’m prone to inverting the Boolean logic compared to shell. If this looks OK to you I’ll mark it ready.

RMerl:
That commit is broken, there's a missing closing paranthesis.

dave14305:
How embarrassing. Added now. I was debating if it should be set unconditionally in the “custom” case.

RMerl:
That's a good question.  I wonder if having an argument specifying the protocol version (4 or 6) would be better.  People with both would just need to call it twice with each version.  In the absence of a protocol version, the default would be both.  I need to see if it's possible to implement while making that new parameter optional, so not to break any existing configuration.

RMerl:
After writing the code and staring at the logic behind it for a few minutes, I don't think having a separate flag for IPv6 makes sense, since you technically cannot update just IPv6 - the internal code doesn't distinguish both as being separate.  So always updating IPv4 but adding the IPv6 flag if ddns support was enabled for IPv6 as your current patch does is what makes the most sense.

PR #895, opened 2026-07-20
feat(vpn-director): add IPv6 policy routing

shedowe19:
## Summary

Adds IPv6 policy routing support to VPN Director for WireGuard and OpenVPN clients.

- Accept IPv4 or IPv6 CIDRs in VPN Director source/destination fields.
- Reject invalid CIDRs and mixed-family rules explicitly.
- Install and remove matching `ip -6 rule` policies alongside IPv4 rules.
- Extend WireGuard route-table handling for IPv6 endpoints and `AllowedIPs`.
- Add an unreachable IPv6 default route in WireGuard policy tables to avoid WAN fallback when no tunnel route applies.
- Extend IPv6 killswitch rules and the HND WireGuard LAN-bypass list.
- Replace predictable temporary route files in the touched WireGuard path with pipe-based reads.

## Compatibility and scope

Existing IPv4 rule format and behaviour are preserved. Empty source/destination rules remain dual-stack and emit one policy rule per enabled IP family. Rules combining IPv4 with IPv6 are skipped because Linux policy rules cannot match both families in one rule.

This intentionally does not introduce IPv6 DNS DNAT for the existing "Exclusive DNS" feature; that currently uses IPv4 NAT semantics and needs separate design work.

## Validation

- Added focused C tests for IPv4/IPv6 CIDR parsing, address-family detection, prefix limits, and mixed-family rejection.
- Built the new helper with `gcc -std=c11 -Wall -Wextra -Werror` and executed the focused tests.
- Executed JavaScript validation tests for the VPN Director form, including valid IPv4/IPv6 CIDRs and invalid prefix lengths.
- Verified `git diff --check`.
- A full router firmware build and target-router runtime test were not available in this environment because they require a selected ASUS platform configuration and its cross-toolchain.

### DIFF
--- a/release/src/router/libovpn/Makefile
+++ b/release/src/router/libovpn/Makefile
@@ -15,7 +15,7 @@ LDFLAGS += -L$(TOP)/openssl -lcrypto -lssl $(if $(RTCONFIG_OPENSSL11),-lpthread,
 
 INSTALL = install
 
-OBJS = openvpn_config.o openvpn_control.o openvpn_setup.o openvpn_options.o amvpn_routing.o openvpn_utils.o
+OBJS = openvpn_config.o openvpn_control.o openvpn_setup.o openvpn_options.o amvpn_routing.o openvpn_utils.o vpndirector_utils.o
 
 all: libovpn.so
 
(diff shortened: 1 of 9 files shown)


RMerl:
Since this is untested (so probably largely vibe-coded) and makes extensive changes all over the place (which would make things difficult to maintain when merging future GPL releases), I am rejecting this.

PR #896, opened 2026-07-23
Fix time-axis tracking in system temperature and 24-hour traffic charts

ExtremeFiretop:
SNB report found here: https://www.snbforums.com/threads/asuswrt-merlin-3006-102-8-is-now-available.97535/post-998653

- Allow the temperature chart’s x-axis to continue advancing after the initial 57-second window is filled.
- Keep the existing 20-sample rolling data window.
- Format elapsed time as `minutes:seconds` to keep longer durations readable.
- Align hourly x-axis grid lines with actual top-of-hour boundaries.
- Use the same time calculation for both x-axis labels and tooltip timestamps.
- Fix cases where a data point with a tooltip such as `18:01` appeared before the `18:00` grid line.

### DIFF
--- a/release/src/router/www/Main_TrafficMonitor_last24.asp
+++ b/release/src/router/www/Main_TrafficMonitor_last24.asp
@@ -83,6 +83,16 @@ const samplesMax = 2880;
 const updateInt = 30;
 const showBitrate = 1;
 
+function chartTime(index) {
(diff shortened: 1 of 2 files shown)


ExtremeFiretop:
Will split this PR into 2 separate commits so review can take less time.

PR #898, opened 2026-08-01
[SECURITY PROPOSAL] Replace OpenSSL 1.1 with  a Compatibility Shim and Upgrade codebase to use OpenSSL 3.5 LTS.

RSDNTWK:
This is a draft proposal for the following changes:

Replace OpenSSL 1.1 codebase for 3.5 LTS.
Upgrade router models using 3.0+ to the same OpenSSL 3.5 version.
Allow vendor prebuilt blobs that require specific 1.x symbols and functions to use an OpenSSL 1.1 <--> OpenSSL 3.5 Shim to remain functional under the 3.5 codebase using only the required functions needed until those components are upgraded or removed by ASUS.

As there is a risk of causing potential compatibility issues. The bare minimum changes have been implemented while allowing easier maintenance for all models. The Shim itself is a compatibility layer that can be updated or removed if changes are required as maintaining the older OpenSSL 1.1 codebase might prove problematic if OpenSSL or Canonical decide to suddenly end support for it in the future.

The OpenSSL 3.5 and Shim have been undergoing live testing using the current codebase on an RT-BE88U and no known issues have been found so far. Other models are untested due to the lack of the hardware required to test them.


RMerl:
I haven't looked at this in details, but the first thing that came up is that you are completely replacing OpenSSL 1.1 rather than using Asus' own existing OpenSSL 3.5 recipes, which they use in Asuswrt 6.0 (with the GT-BE19000AI being the only model currently using it).

A cleaner approach would be to use the same recipes as used in Asuswrt 6.0, leaving OpenSSL 1.1 in place.  If the build recipe must be adjusted for non-Asuswrt 6.0 devices, one possible way could be to base it on model name (since so far GT-BE19000AI being the only model with full native 3.5 support).  Another possibility would be to add a new target.mak flag stating if the shim need to be used/enabled.

Regarding Hosttools: I don't think Asuswrt uses the openssl archive present there, so I would skip changing that, it only adds more dead weight to the already oversized git blob repo.

Ultimately, I'm not sure if I'd be confident enough in going forward with this, due to the high potential of causing compatibility issues within the binary blobs that are linked and tested against 1.1.  This might make troubleshooting issues more difficult, as there will always be a risk that the issue might be related to this forced change.

I wish Asus simply went ahead and migrating every model to 3.5, since they already got the code working for Asuswrt 6.0 devices.  Unfortunately at this time they don't seem willing to make the transition, for some unknown reason.  I might need to start putting more pressure on them on this front.

RSDNTWK:
@RMerl Fair enough. It was a suggestion due to fact that if and when Asus would be updating their binary blobs to support the newer OpenSSL versions and the unknown factor of how long OpenSSL or Canonical will continue supporting extended updates for OpenSSL 1.1.1. The Shim itself is designed to only translate the required functions to the newer OpenSSL versions without requiring the actual library version itself as the blobs would see it as the older version by default.

RMerl:
Ubuntu 20.04 was the last LTS that used OpenSSL 1.1.1.  Official support ended last year, but they still offer paid-for support until 2032.  Whether it will remain easy to get access to those patches during the paid-for support window is unknown.

RSDNTWK:
@RMerl The same can be said for OpenSSL as well as they have not disclosed the length of time they will continue supporting it either, at least not publicly.

RSDNTWK:
Unless you wish to close this, this is just an idea to make the repo more maintainable unless Asus start implementing newer blobs to avoid this sort of thing.

RMerl:
Let me check with them what their current plans are.  If it's a negative on their end, then we can discuss how best to implement this for at least the rest of the 3006 models.  As indicated above, my recommendations would be to stick as closely as possible to Asus's implementation in Asuswrt 6.0 (meaning no override of the existing 1.1 code), with the only change being inserting support for the shim for "non-native" models.

RMerl:
Unfortunately, Asus seems unwilling to commit to any timetable or plan for the migration of Asuswrt 5.0 models to OpenSSL 3.5.

I had a quick review of your commit, and I believe it could be made far less intrusive by being based on Asus' own 3.5 implementation.  For instance instead of modifying every Makefile to refer to the opensl-3.5 folder through a var, it simply refers to the openssl-3.5 folder through a symlink pointing router/openssl/ to router/openssl-3.5.  That way, when RTCONFIG_OPENSSL11 is not defined, the symlink takes care of pointing at the correct 3.5 location.

I recommend you do a test compile of the asuswrt6 branch so you can look at the result, and see if you can reproduce a shim-based implementation by staying closer to Asus' implementation.  If it can be done without making extensive changes to the rest of the firmware, I might be willing to give this a try.

Unsure if you'd rather start from a brand new PR or modify this one - up to you.

RSDNTWK:
The OpenSSL 1.1 vendor blob dependency list is pretty small as everything else will compile with Openssl 3.5:

release/src/router/amas-utils
release/src/router/libasc
release/src/router/libletsencrypt

Libraries like rc, httpd, and infosvr would only need to have the shim added on top based on the asuswrt6 tests so far.

RSDNTWK:
@RMerl Closing this and created #901 in place of it.

PR #900, opened 2026-08-01
webui: Remove temperature graph x-axis labels

dave14305:
I propose to revert the recent changes by @ExtremeFiretop and just remove the labels completely. The increasing mm:ss labels don't carry much meaning, especially in a graph only covering a 60-second window.

The Traffic Monitor real-time graphs do not have x-axis labels either.

Open to suggestions, arguments, etc.

### DIFF
--- a/release/src/router/www/Tools_Sysinfo.asp
+++ b/release/src/router/www/Tools_Sysinfo.asp
@@ -148,7 +148,7 @@ var wifi51data = [];
 var wifi52data = [];
 var wifi6data = [];
 var wifi62data = [];
-var tempchartOffset = 0;
+
 
 function draw_mem_charts(){
 /* Memory */
@@ -278,13 +278,6 @@ function draw_mem_charts(){
 
 
 function draw_temps_charts(){
-	if (cpudata.length > 20 ||
-	    wifi24data.length > 20 ||
-	    wifi51data.length > 20 ||
-	    wifi52data.length > 20 ||
-	    wifi6data.length > 20 ||
-	    wifi62data.length > 20)
-		tempchartOffset += 3;
 	if (cpudata.length > 20)
 		cpudata.shift();
 	if (wifi24data.length > 20)
@@ -418,15 +411,6 @@ function draw_temps_charts(){
 					labels: [0,3,6,9,12,15,18,21,24,27,30,33,36,39,42,45,48,51,54,57],
 					ticks: {
 						color: ticksColor,
-						callback: function(value) {
-							var seconds = Number(this.getLabelForValue(value)) + tempchartOffset;
-
-							if (seconds < 60)
-								return seconds;
-
-							return Math.floor(seconds / 60) + ":" +
-								String(seconds % 60).padStart(2, "0");
-						}
 					}
 				},
 				y: {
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
I'm fine either way. I just couldn't stand the static values.
I would be open to hearing @RMerl 's opinions on the graph for sure

dave14305:
Before:
image

After:
image

RMerl:
I experimented with a few alternatives before merging the last commit, and none were any better.  Not having anything on the X scale means you will have no sense of the time scale between the start and the end of the chart, while having a numbered scale tells you that the chart shows up to 60 seconds.

dave14305:
This was an earlier option to show the relative time and start populating from the right-hand side (zero seconds ago).

RMerl:
Just to compare, any idea how the rest of the market (TP-Link/Netgear/pfsense/Ubiquiti/etc...) handle these types of real time graphs?

ExtremeFiretop:
> Just to compare, any idea how the rest of the market (TP-Link/Netgear/pfsense/Ubiquiti/etc...) handle these types of real time graphs?

I only have access to one Friends TP-LINK Omada Mesh system, but for reference, it doesn't even have historical data, just "live" data:

Screenshot 2026-08-02 235907

Screenshot 2026-08-02 235925

ExtremeFiretop:
I also tried to poke more at this, unfortunantly this is as good as I got:

image

image

image

eoleedi:
> Just to compare, any idea how the rest of the market (TP-Link/Netgear/pfsense/Ubiquiti/etc...) handle these types of real time graphs?

Those are real-time graphs and history graphs of ubiquiti's internet traffic.

Screenshot 2026-08-03 at 12 25 16
Screenshot 2026-08-03 at 12 24 57

ExtremeFiretop:
> > Just to compare, any idea how the rest of the market (TP-Link/Netgear/pfsense/Ubiquiti/etc...) handle these types of real time graphs?
> 
> Those are real-time graphs and history graphs of ubiquiti's internet traffic.
> 
> Screenshot 2026-08-03 at 12 25 16 Screenshot 2026-08-03 at 12 24 57

What about CPU and Radio temperatures?
That's what we are currently looking at. Does ubiquiti have something to compare too?

eoleedi:
> > > Just to compare, any idea how the rest of the market (TP-Link/Netgear/pfsense/Ubiquiti/etc...) handle these types of real time graphs?
> > 
> > 
> > Those are real-time graphs and history graphs of ubiquiti's internet traffic.
> > Screenshot 2026-08-03 at 12 25 16 Screenshot 2026-08-03 at 12 24 57
> 
> What about CPU and Radio temperatures? That's what we are currently looking at. Does ubiquiti have something to compare too?

I managed to find some temperature graphs online on unifi environmental sensor, but overall it is the same.
Ubiquiti doesn't seem to show real-time temperature of CPU or radio sensors, at least on my devices.

Screenshot 2026-08-03 at 15 15 41

RMerl:
I suspect a lot of platforms will just show historical data rather than realtime charting like Asuswrt-Merlin does.

I think that at this point, this is a case where we might all be overthinking it, with everyone having a preferred implementation, without any of them being clearly superior - it mostly boils down to personal preferences.  Personally I was equally fine with both the old and the current implementation.   Seeing as nobody outside us three have discussed this ever since it was recently changed, I suspect that most end users just don't care - as long it's clear that the whole chart covers 60 seconds, and they see it changing every few seconds.

I'd be inclined for the time being to simply keep things as they are, so not to confuse users by changing the charts for a second time over the course of two releases.

PR #901, opened 2026-08-05
OpenSSL 3.5 + Shim Implementation.

RSDNTWK:
Rewrite of #898 with minimal changes as requested.

Adopted asuswrt6 branch recipes for OpenSSL 3.5.

OpenSSL 1.1.1 dependent vendor blobs below adopt the minimal Shim for compatibility:

release/src/router/amas-utils
release/src/router/libasc
release/src/router/libletsencrypt

rc, httpd, infosvr and libcreduction adjusted to incorporate the Shim's libcrypto.so.1.1 and libssl.so.1.1 libraries for the above dependencies. 

The Shim substitutes OpenSSL 1.1.1 itself and handles the bare minimum translation handling necessary to OpenSSL 3.5.


RMerl:
Don't remove OpenSSL 1.1 from the tree.  It's still a valid build option, and may be required at any time for test purposes if anything breaks with the 3.5 switch.

RSDNTWK:
@RMerl Amended the previous commit to restore openssl-1.1 folder apologies about that.

RMerl:
No worries, thank you for the amendment.

What provides the openssl11-compat that is linked with?

RSDNTWK:
@RMerl openssl11-compat is the shim that directly wraps only the required OpenSSL 1.x symbols used by amas-utils, libasc and libletsencrypt directly to OpenSSL 3.5 instructions. So essentially when they are used by the vendor binaries, OpenSSL 3.5 directly handles the symbols functions through it's own library version via the shim's own libcrypto.so.1.1 and libssl.so.1.1 wrapper libraries in place of OpenSSL 1.1 itself. An example is like this:  libletsencrypt <--> openssl11-compat <--> openssl-3.5. Since rc, httpd, infosvr and libcreduction call those 3 libraries that use OpenSSL 1.1, the shim was added to allow the symbols to correctly link and resolve to the shim when compiling.

RMerl:
I understand that.  What I mean is, what generates the shim file?

RSDNTWK:
It has a simple makefile and uses both the compat c files and the symbol maps to create the wrapper's libcrypto.so.1.1 and libssl.so.1.1 while it links against OpenSSL 3.5 and compiles it to the targets/<profile>/fs.install/usr/lib/ directory.

RMerl:
Should there be a router/openssl11-compat/ folder then?  I don't see it in the commit (hence my confusion). :)

RSDNTWK:
@RMerl It might not show due to the limit of the number of files shown on the "Files Changed" tab being capped at 3,000 files shown. It is committed here: https://github.com/RSDNTWK/asuswrt-merlin.ng/tree/shim-rewrite/release/src/router/openssl11-compat

RMerl:
Gotcha.  I was able to retrieve the whole patch from SSH:

```
wget https://patch-diff.githubusercontent.com/raw/RMerl/asuswrt-merlin.ng/pull/901.patch
```
That will allow me to review + test it, thanks.

RMerl:
I did a quick test build on a GT-BE98_Pro, things seem ok so far.

I'm still a bit reluctant on relying on a shim for a core component such as OpenSSL, but I'm willing to keep this option open in case Asus takes too long in migrating other routers or ultimately decides not to do it, or to be convinced that it would be a good thing to do.  I had a few questions:

1) What is the origin of the shim?  Did you write it yourself, or is it based on already existing code from elsewhere?
2) What is the biggest risk in terms of reliability.  Any chance, for example, that a remapped call might break something?
3) What would happen if, for example, a remapped call was missing, following a future change from Asus in their closed source portions?  Would the build simply fail as it would try to link to a non-remapped API call, or would the build succeed but that code path misbehave?

RSDNTWK:
It was written from scratch while referencing the functions from both openssl versions needed while testing with the BE88U. The remapped calls would only break if a new change needed updating for example if openssl 3 changed the way it worked or produce an undefined reference error if a new function is missing
that's required later. It's intended to wrap calls it needs to work with and it fails for ones it doesn't have to minimise security risk exposure to.

RMerl:
I'm currently adapting the merge for the 3006.102-wifi6 branch (which uses a slightly older 3006.102 codebase than for WIfi 7 models).  Once done, the current plan is to generate test images for all 3006.102 devices, and have the community test them for a while.

If that goes well enough, I'll see if it can also be adapted to the 3004.388 models.  I generally try to be conservative with changes for that legacy branch, but if I end up switching to OpenSSL 3.5 through a shim, the goal is for long-term support regardless of when OpenSSL 1.1.1 support might disappear.  That goal would make this change make sense for 3004.388 devices where the current plan is extended support.

RMerl:
Public testing started:

https://www.snbforums.com/threads/openssl-3-5-firmware-test-builds.97713/

RSDNTWK:
Everything should work ok based on the BE88U live testing performed so far internally here. In the mean time any backports for 1.1.1 will continue if any become available.

RSDNTWK:
@RMerl The forum posts seem to report everything is working as expected so far. Also double checked the shim by running a letsencrypt cert generation which worked perfectly against the known libletsencrypt.so being hardcoded to 1.1.1.

RMerl:
Looking good indeed.  I will most likely be going ahead with this in the next major release (3006.102.9).

One thing I need to look at is EasyRSA.  Asus switched to version 3 on the GT-BE19000AI at the same time they moved to OpenSSL 3.x,  I need to review the other code branches that are still on version 2.  I need to check if I can safely upgrade to 3, or if there are closed source components that expect version 2 (the API is a bit different)  I`m not entirely sure if that older version 2 would work properly with OpenSSL 3.5.

Later on I'd also need to check if it would be doable to port all of this to the older 3004.388 branch.  That branch lacks any 3.5 support, so in addition to the build recipes, some other parts may need to be adjusted for 3.5 support, ensuring the right OpenSSL includes are used, etc...

RMerl:
I did a grep for "easy-rsa" through the whole router/ folder, and the only place that it's currently in use is within libovpn.  So going forward with switching to EasyRSA 3 will be fine, I'll port my changes from the asuswrt6 branch.

RSDNTWK:
@RMerl Just did a quick compile test on the 3004 branch using the RT-AX88U profile. With some minor adjustments the changes worked as is with the firmware code. Unable to test functionality however.

RMerl:
I was expecting a few Makefile changes to components like httpd, but I suspect the router/openssl symlink is making everything automatically point at the OpenSSL 3.5 includes.  I'll try cherry-picking + adjusting this commit on the 3004.388 branch later on, I'm currently focusing on bugfixes ahead of 3004.388.12_2 and 3006.102.8_2 point releases, which will only include bugfixes.

I updated libovpn on the main branch to use EasyRSA 3 when under OpenSSL 3.0/3.5, so that part should be easy for me to apply later on to the 3004.388 branch as well.

RMerl:
When you get the chance, can you provide a separate PR for the 3004.388 branch?  I will take care of getting actual functionality tested afterward.

RSDNTWK:
@RMerl Done, see PR #902.

RMerl:
The RT-BE58_Go branch will require some investigating, build fails while running libcreduction.

```
Makefile:220: *** libcreduction FATAL: Missing 32-bit libraries: libssl.so.1.0.0.  Stop.
```

Probably a Makefile that thinks it's using OpenSSL 1.0 instead of 3.5 due to a missing/incorrect #ifdef .

RSDNTWK:
@RMerl Most likely a minor change will be needed, looking into that now.

RSDNTWK:
@RMerl Fixed the Compile error in #904 and also updated the shim with additional symbols needed in #905 that can be merged across all branches.

RMerl:
Thank you.  I will test this later tonight, and also run some tests on the HNS functionalities to ensure that part is still working as expected.

RMerl:
The ufwd userspace daemon seems to be running, but it's hard to do any real test beyond that since the HNS licence used on the RT-BE58GO is limited to handling the IPS, and I never found an actual test that could trigger an IPS event.

We'll have to see how things go once it goes into public testing.

Thank you for the quick fix.

RSDNTWK:
@RMerl No problem, one discovery found is hns still relies on the older OpenSSL 1.0.x symbols. A specific RT-BE58_Go 1.0.x shim is possible to do as well tieing it to the same openssl 3.5 library,.

RMerl:
If you want to keep the shim slimmer and avoid having a separate shim version for the BE58GO, you could enclose the relevant code sections into RTCONFIG_HNS define blocks, as these are really for HNS, not the BE58GO itself.  I doubt I'll ever support any other HNS-based model, but who knows.

RSDNTWK:
@RMerl Currently looking into an HNS only openssl10-compat shim. Seems HNS uses 53 symbols for it and managed to get built the same way. it will only compile if both RTCONFIG_HNS=y and RTCONFIG_OPENSSL35=y.

RMerl:
It's just insane that Trend Micro, a friggin' security company, is still releasing software relying on such an antique, insecure version.  1.0.2 went EOL in 2019, 7 years ago.

If supporting it properly is going to prove to be too problematic, I'm fine with leaving the RT-BE58_Go on OpenSSL 1.1.1.

RSDNTWK:
@RMerl Not that hard as a lot of the symbols needed the 1.1 shim already has, the adjustments made so far on the extra shim are 1.0.x specific changes that are separately needed but could be merged into the 1.1 shim later to unify it.

RSDNTWK:
@RMerl Added an OpenSSL 1.0.x shim in #908 as it was safer than merging them due to the differences in the symbol implementations. Qemu arm tests confirm linkage and basic binary testing. Unable to test on actual hardware though but should work.

